PR #8, opened 2017-01-25
Add some tests and fix tc.cc

mtorpey:
Just a few tests by @ChristopherRussell and me, as well as a fix for Todd-Coxeter.


james-d-mitchell:
looks good @mtorpey but can you please change the commit prefixes to `cong` and `tests` respectively?

PR #9, opened 2017-01-25
Fix issues with non-reduced RWSEs

wilfwilson:
I've added some tests for finite fp-semigroups which have some generators that are equal. Currently Todd-Coxeter deals with these fine, but KBP and KBFP seem to assume that these generators are distinct, even though one can be reduced to the other by the rewriting system. In order words, Knuth-Bendix is currently assuming that generators are automatically reduced.

Note that these tests currently fail for KBP and KBFP.


james-d-mitchell:
@wilfwilson happy to merge this, but can you please change the prefix of the second commit to `tests`?

wilfwilson:
@james-d-mitchell I've made the change (thankfully the checks still passed!).

PR #10, opened 2017-02-10
Change nontrivial_classes to use words instead of element pointers

mtorpey:
In order for `Congruence::nontrivial_classes` to have a meaningful output in the case of finitely presented semigroups, it must return words in the semigroup's generators, rather than element pointers as it does now for enumerable semigroups.  Hence this pull request modifies the definition of `partition_t` from `vector<vector<Element*>>` to `vector<vector<word_t>>`, and alters all the `nontrivial_classes` functions in the various `DATA` subclasses as appropriate.

One additional feature is required in order to make this happen: a way of directly extracting an `RWSE`'s underlying word.

Why is this necessary?
* If we have a `Semigroup*` and an `Element*`, we normally get a `word_t` describing the element by calling `_semigroup->factorisation(w, pos)`.
* To get the `pos` for an element, we have to call `_semigroup->position(elm)` which calls `Semigroup::enumerate`.
* If the semigroup is infinite, `enumerate` may run forever without finding the element, or may find it in unbounded time.
* However, in all infinite cases the `Element*` will actually be an `RWSE*`, which already knows the factorisation of the word (in fact the element is defined by an `rws_word_t`).
* So, we should extract this underlying word wherever necessary in `nontrivial_classes`, instead of using `Semigroup::factorisation`.

The first commit in this PR simply adds a static function `rws_word_to_word` which takes an `rws_word_t` and returns the corresponding `word_t`.

The question is how we should go about extracting that word.  Since `Congruence` functions do not know the type of the `Element*`, we cannot have something that looks like
```
if IsRWSE(elm) {
  // use the underlying word
} else {
  // use Semigroup::factorisation
}
```

The second commit in this PR contains my suggested solution to the problem.  We add a property to the `Element` class, called `has_easy_word_representation`.  This is a bool, and the virtual function installed in the `Element` class simply returns `false`.  Now every other subclass of `Element` can safely ignore all of this, but any classes which do have an underlying word which can be extracted easily, like `RWSE`, can override it with `true`.

Now functions that require a `word_t` to represent an element can check `has_easy_word_representation`, and *only if it returns true*, they can then call the new function `Element::as_word_t`.  A dummy `as_word_t` function is installed in the `Element` class, and should never be called.  The `RWSE` class overrides this with a function that returns the `word_t` representation of its underlying `rws_word_t`.

Clearly, `as_word_t` should never be called unless `has_easy_word_representation` is satisfied.

Now that this is set up, the third commit in this PR alters all the `nontrivial_classes` functions to handle `word_t`s instead of `Element*`s.  The only place that `as_word_t` and `has_easy_word_representation` are used is in `P::nontrivial_classes`.

After this is merged into libsemigroups, I have a branch for the Semigroups package which makes the whole congpairs ecosystem work with these changes, thus allowing FP semigroups to use our code, and fixing that annoying test in `semicong.tst`.


mtorpey:
I've redone this pull request, and it's now what @james-d-mitchell and I discussed.  We have a new enum called `elm_t` which indicates whether an element is `RWSE` or `NOT_RWSE`, and we can use it to avoid doing costly (and infinite?) enumeration when we factorise an element, if we already have a factorisation easily available.

Note that due to rebasing, the commits are displayed on this page in a funny order.  They're actually on the branch in a logical order, and each one contains a sensible discrete piece of work, so there should be no need for squashing.

elements.h
```diff
@@ -43,10 +43,21 @@ namespace libsemigroups {
...
+    enum elm_t {
+      RWSE = 0,
+      NOT_RWSE = 1
+    };
+
     // 0 or 1 parameter (hash value)
     // @hv the hash value (for caching) of the element being created (defaults
     // to <libsemigroups::Element::UNDEFINED>.
-    explicit Element(size_t hv = Element::UNDEFINED) : _hash_value(hv) {}
+    explicit Element(elm_t type, size_t hv = Element::UNDEFINED)
```
james-d-mitchell: The `explicit` should be dropped.

cong/tc.cc
```diff
@@ -611,41 +611,29 @@ namespace libsemigroups {
...
     //       because the classes are not numbered contiguously {0 .. n-1}
-    std::vector<std::vector<size_t>> pos_classes(_table.nr_rows(),
-                                                 std::vector<size_t>());
+    partition_t all_classes(_table.nr_rows(), class_t());
     // Look up each semigroup element's position in the table
     for (size_t pos = 0; pos < _cong._semigroup->size(); pos++) {
       _cong._semigroup->factorisation(word, pos);
       assert(word_to_class_index(word) < _table.nr_rows());
-      pos_classes[word_to_class_index(word)].push_back(pos);
+      all_classes[word_to_class_index(word)].push_back(word);
```
james-d-mitchell: Is this actually valid? In particular, are you certain that `word` is copied when it is pushed into `all_classes`? Given that `factorisation` changes `word` in-place, are you sure that you don't end up with a vector of identical words? I'd like to see tests verifying that this is not the case.

cong/tc.cc
```diff
@@ -611,41 +611,29 @@ namespace libsemigroups {
...
     word_t word;
     // TODO: This may create many empty vectors that won't be used,
     //       because the classes are not numbered contiguously {0 .. n-1}
-    std::vector<std::vector<size_t>> pos_classes(_table.nr_rows(),
-                                                 std::vector<size_t>());
+    partition_t all_classes(_table.nr_rows(), class_t());
     // Look up each semigroup element's position in the table
     for (size_t pos = 0; pos < _cong._semigroup->size(); pos++) {
       _cong._semigroup->factorisation(word, pos);
       assert(word_to_class_index(word) < _table.nr_rows());
```
james-d-mitchell: Is this check meant to verify that the `class_index` is within the correct range? Do we know at this point that the class indices are in a range `{0, ..., n - 1}`?

mtorpey: At this point in TC we cannot rely on the indices being in a range `{0, ..., n - 1}` because we have not compressed the table.  Hence, we can only rely on it being between zero and the number of rows in the table.  The corresponding code in `kbfp.cc:100`, where we *can* rely on the range `{0, ..., n - 1}`, has an assert that checks the index is less than `nr_classes()`.

elements.h
mtorpey: Will do.

partition.h
```diff
@@ -19,13 +19,17 @@
...
+
 namespace libsemigroups {
 
   // Non-abstract
   // Class for representing a partition of a semigroup, i.e. an equivalence
-  // relation on some subset of the semigroup.
+  // relation on some subset of the semigroup.  T is the type of element stored,
+  // and may be a pointer type, e.g. word_t*
   //
   template <class T> class Partition {
```
james-d-mitchell: I know this is WIP, but shouldn't it be `template <typename T>`?

mtorpey: I looked this up and I understand the two are equivalent.  But perhaps `typename` would be better, since it doesn't imply that we're using a class (e.g. in the main use-case, `Partition<word_t*>`)

mtorpey:
I believe this is ready for merging now.

mtorpey:
I did a little valgrinding, and added two lines to make sure there are no leaks.  I'm now quite sure this is ready for merging.

For anyone reading this who isn't @james-d-mitchell, this PR now adds a class called `Partition` which wraps up a `vector<vector<word_t*>*>*` describing an equivalence relation on some subset of a semigroup.  This is used for `nontrivial_classes`.

james-d-mitchell:
This is generally looking ok. But @mtorpey can you please rebase this onto the top of the master branch? 

As it is I can't run the tests for code coverage. Also, I guess you didn't compile the documentation, please do this, and fix the issues you'll see, in particular, look at the `Documentation generator` which will point out some things that are missed.

mtorpey:
I've rebased onto master now - I was initially reluctant to base a pull request on top of a commit with "TEMP" in the title.

As for the doc, after multiple hours trying to configure cldoc on my two different computers, I've finally decided I have to give up - something about clang 4.0 and my system's libraries just aren't playing together, and the problem seems to match a "known issue" on the cldoc website.  On the plus side, it looks like I might now be able to use clang-format!

Would it be possible for you to post the output of `make doc` here, so I can see what's going wrong and attempt to fix the documentation problem?

wilfwilson:
I've never managed to get the documentation to compile either @mtorpey

james-d-mitchell:
Hmm, I've never had any problems. I think I'd like to stop using cldoc because it doesn't seem to be being developed any longer. In the mean time, I've posted the output from your branch at: [https://goo.gl/0lNvFS](https://goo.gl/0lNvFS). It will be hard for you to fix everything without being able to compile the doc, so just try to do as much as you can, and I'll take it from there.

mtorpey:
Thanks for the report.  I've added all the missing doc I could find that was related to this pull request - hopefully not too much more needs done.

Glad to hear we're moving away from cldoc - it's very much tied in with the whole LLVM system, and seems to rely on more recent versions of clang than are stable in Debian.  I feel like I learned a lot about compilers yesterday!

PR #17, opened 2017-03-24
Batch processing for congruences

mtorpey:
This PR allows a congruence to be run "a bit at a time", allowing us to answer whether two elements are in the same class without running to completion.  In particular, this allows us to say whether two elements are related even for congruences with an infinite number of pairs.

This has required quite a lot of changes, since we need to do some things quite differently.  When reviewing, I recommend looking through the changes one commit at a time, from the start - this should make it a bit more understandable.

### DIFF
--- a/cong.cc
+++ b/cong.cc
@@ -225,42 +225,44 @@ namespace libsemigroups {
     return winner;
   }
 
-  void Congruence::force_tc() {
+  void Congruence::clear_data() {
     if (_data != nullptr) {
       delete _data;
     }
+    if (!_partial_data.empty()) {
+      for (size_t i = 0; i < _partial_data.size(); i++) {
+        delete _partial_data.at(i);
+      }
+      _partial_data.clear();
+    }
+  }
+
+  void Congruence::force_tc() {
+    clear_data();
     _data = new TC(*this);
   }
 
   void Congruence::force_tc_prefill() {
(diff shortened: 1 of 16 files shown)


wilfwilson:
Do the commits on the commits page appear in the correct chronological order? Because I know GitHub likes to screw with the ordering sometimes, which can be confusing...

james-d-mitchell:
Looks like they have the original dates after some rebasing.

mtorpey:
I've made all the changes we discussed, so I believe this is ready for merging.

mtorpey:
I've made all the changes we discussed, so I believe this is ready for merging if you're happy with it.

mtorpey:
I've made all the changes we suggested, including a `result_t` enum for unknown values, and `test_less_than` etc.  All the changes I've made to non-cong files are at the start, including `RWS::test_less_than`.  I think this is ready for merging (but I'm in the final stages of plumbing it into GAP).

mtorpey:
New feature: we now keep old incomplete `DATA` objects in memory, only destroying them if one reaches an `is_done()` state.  They're stored in a new vector called `_partial_data` and only placed in `_data` when completely enumerated (or when created by a `force_x` method).

This change is pretty much concentrated in the last three commits, if you want a good idea of what happens.

About to push.

PR #20, opened 2017-04-12
Make nontrivial_classes use KBP for an fp congruence

mtorpey:
KBP is the only `DATA` which gives a meaningful answer for `nontrivial_classes`.  Hence, we should use KBP whenever `nontrivial-classes` is asked for.

This fixes @james-d-mitchell's segfaulting test.

### DIFF
--- a/cong.cc
+++ b/cong.cc
@@ -266,6 +266,26 @@ namespace libsemigroups {
     _data = new KBFP(*this);
   }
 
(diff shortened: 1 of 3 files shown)


james-d-mitchell:
@mtorpey please run `make format` in the directory, there is one formatting issue, which this will resolve. Otherwise, I'm happy to merge.

mtorpey:
@james-d-mitchell Done.

PR #23, opened 2017-04-16
Fix TC to allow a non-empty coincidence stack

mtorpey:
Just a little thing I didn't spot now we're allowing `DATA` objects to be resumed after being killed.  In TC, the coincidence stack might have a few pairs on it when killed.  We can process them when we resume, so this PR removes the assertion that the stack is empty (as well as adding a few more asserts for good measure).

### DIFF
--- a/cong/tc.cc
+++ b/cong/tc.cc
@@ -372,7 +372,8 @@ namespace libsemigroups {
   void Congruence::TC::identify_cosets(class_index_t lhs, class_index_t rhs) {
     TC_KILLED
 
-    assert(_lhs_stack.empty() && _rhs_stack.empty());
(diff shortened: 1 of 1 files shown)


james-d-mitchell:
Interesting, I saw the tests fail a couple of times because of this. Good you caught it!

james-d-mitchell:
This is a more or less serious issue, right? I should make another release with these changes?

mtorpey:
A release with these changes would make sense in my opinion, since this could be the difference between a crash and no crash.  I assume it's fairly easy to make a release?

james-d-mitchell:
Yes, pretty easy. But then this should be merged into `stable-0.1`

PR #28, opened 2017-04-20
tc: in init, only set _tc_done to true if not killed

mtorpey:
This PR fixes the issue exposed by commit `668d6ea`, which is causing [the current Semigroups build to fail](https://travis-ci.org/gap-packages/Semigroups/jobs/223708980).

If a TC object was killed while prefilling, it might have empty relations during `init()`.  But this doesn't mean it's done.  We should only allow it to set _tc_done to true if it hasn't already been killed.

I also removed a FIXME: even if we are prefilled here, we still always need to apply the "extra" relations.

### DIFF
--- a/src/cong/tc.cc
+++ b/src/cong/tc.cc
@@ -104,12 +104,10 @@ namespace libsemigroups {
       // This is the first run
       init_tc_relations();
       // Apply each "extra" relation to the first coset only
-      // FIXME couldn't we be prefilled here, in which case the next line
-      // performs a potentially costly computation
       for (relation_t const& rel : _extra) {
         trace(_id_coset, rel);  // Allow new cosets
       }
-      if (_relations.empty()) {
+      if (_relations.empty() && !_killed) {
         _tc_done = true;
         compress();
         return;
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
Good catch. Are you completely sure that we can't be killed in the line of code this PR modifies, and be done, in a situation that everything ends up killed, and this is the only thread that is actually done but won't know it because of this change? Does that make sense?

Also the FIXME was about when `init()` should be called. In the `current_equals` method the first line was `init()` which could be costly, then in the next line if `(!is_done() && is_killed())` then we stop. I reordered this to just check `(!is_done() && is_killed())` first (which is cheap) and then run `init()` because the rest of the method should be valid even if the object is killed in the call to `init()`.

james-d-mitchell:
Also: can you please add at least one test to `libsemigroups` that highlights the issue that this fixes?

james-d-mitchell:
And this should be merged into `stable-0.2` when ready

mtorpey:
I've added a test for this (in fact it's the one that was failing in Semigroups) and verified that it does indeed fail when run on the master branch.  It passes with this change.

And to answer your question, I am sure this change is okay in terms of recognising success.  If this thread has been killed by another thread, then we're happy to terminate this thread without recognising that we're done - and indeed, we'll recognise it immediately if we resume this thread later.  And there's no way this thread could kill the others without first setting `tc_done` to true.  So I think we're safe.

PR #30, opened 2017-04-21
Add support for finitely presented semigroups

le27:
Introduces `FpSemigroup`, `FpMonoid` and `FpSemigroupElement` and `FpMonoidElement` classes as well as unittests and documentation for all of these.

Note that there is currently no normal form function for words so multiplication is currently done by returning the word to class index of the product.

The following member functions are included:

### FpSemigroup
* `init`
* `size`
* `is_finite`
* `set_report`
* `set_max_threads`
* `alphabet`
* `is_confluent`
* `word_to_class_index`
* `dealloc`
* `repr`

### FpMonoid
* `init`
* `alphabet`
* `repr`

### FpSemigroupElement
* `init`
* `semigroup`
* `word`
* `mul`
* `repr`
* `dealloc`
* `word_to_class_index`

### FpMonoid
* `init`
* `monoid`
* `repr`

### DIFF
--- a/python-bindings/semigroups.pyx
+++ b/python-bindings/semigroups.pyx
@@ -744,6 +744,8 @@ cdef class FpSemigroup(Semigroup):
         for i in xrange(len(word)):
             word[i]=self.alphabet().index(word[i])
 
+
+
         cdef vector[uint64_t] Word = word
         return self._congruence.word_to_class_index(Word)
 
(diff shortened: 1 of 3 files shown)


james-d-mitchell:
The obvious first comment is that `FpMonoid` appears twice in your list of things in this pull request. I guess the second one should be `FpMonoidElement`. The other thing that stands out (before looking at the code) is that `FpMonoid` and `FpSemigroup` do not have the same methods, I can't see why they wouldn't? Perhaps this is a mistake?

james-d-mitchell:
I guess there are several issues with this but I'm going to merge it and then clean them up.

PR #31, opened 2017-04-27
Divide Travis tests into suites, and add new features

mtorpey:
This PR modifies the Travis setup to provide six different suites, each checking a different aspect of libsemigroups:

* `standard`: The tests provided by `make check-standard` (assertions on/off) and `make distcheck`
* `lint`: cpplint is run using `make lint`
* `doc`: Compiles the documentation using doxygen, and fails if there are any warnings (which there are)
* `coverage`: Code coverage tests which require 85% coverage of all `.h`/`.cc` files except `report.h` and `timer.h` (we could change this to 95% if we wanted)
* `valgrind`: Runs valgrind on all tests labeled "quick", and fails if the word "invalid" appears or any bytes are lost (lots of things going wrong here - not sure whether they need to be fixed or whether I should be filtering the output in some way)
* `gap`: Compiles GAP (master) with the Semigroups package (master) and this version of libsemigroups.  Runs `travis-test.sh` from inside Semigroups, without any special features turned on (so just standard tests and manual examples)

The only two which are currently failing are `doc` and `valgrind`, which I believe is correct.  In testing I deliberately sabotaged most of the other tests and I'm pretty confident they'll fail when they're supposed to.

Note that a given suite can be turned off easily by just commenting out the appropriate line in `.travis.yml`


james-d-mitchell:
Awesome!

james-d-mitchell:
Can you try removing the two badges from the top of the README and see if the doc tests work? @mtorpey

mtorpey:
Looks like the badges in README had nothing to do with it - the message given is still the same:
> Resolving user defined `refer/home/travis/build/james-d-mitchell/libsemigroups/src/rws.h:111`: warning: `Member _letter_order(letter_order)` (function) of class `libsemigroups::SHORTLEX` is not documented.

james-d-mitchell:
Right, the version of Doxygen installed in the test is too old, and gives false warnings which are not present with the most recent version.

james-d-mitchell:
Hang on I take it back it looks like the right version, but I dont get that warning on my computer. It's warning that a private data member had no doc, hmmm

mtorpey:
I've fixed the things I talked about with @james-d-mitchell - namely: Valgrind now greps for more stuff and only runs in one session; and Doxygen is updated to 1.8.11 (as recent as I could get that actually installs and runs) and now only fails on two lines:
```
warning: ignoring unsupported tag `TOC_INCLUDE_HEADINGS   =' at line 313, file Doxyfile
warning: ignoring unsupported tag `PLANTUML_CFG_FILE      =' at line 2409, file Doxyfile
```
I could suppress this by ignoring any warnings with "ignoring" in them, but I don't know if this is something that should/could be fixed instead.

james-d-mitchell:
@mtorpey must be that these tags are added in a later version of Doxygen, I guess we are not using them, so maybe we can just comment them out of the `Doxyfile`?

PR #34, opened 2017-05-26
test_less_than should delete all non-winning DATA objects

mtorpey:
There are various changes here, but the important one is "cong: in test_less_than, delete all but the winning DATA".  This ensures that if sorting is ever done, it is done consistently from that time on.  It's possible we'll want to change this later on to always use Knuth-Bendix, or something like that - but for now this makes things consistent.

We've also got a handful of other changes, which should make sure certain bugs never arise.  Each "fix" is in its own commit.

### DIFF
--- a/src/cong.cc
+++ b/src/cong.cc
@@ -209,19 +209,16 @@ namespace libsemigroups {
         winner->run();
         assert(winner->is_done());
       } else {  // Congruence is defined over an fp semigroup
+        std::vector<DATA*>                      data  = {new KBP(*this)};
+        std::vector<std::function<void(DATA*)>> funcs = {};
         if (_type == TWOSIDED) {
-          std::vector<DATA*> data = {new TC(*this), new KBFP(*this)};
-          if (!_relations.empty()) {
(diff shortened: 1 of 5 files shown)


src/cong.cc
```diff
@@ -241,7 +240,60 @@ namespace libsemigroups {
...
+      return true;
+    }
+
+    // Does there exist a generator which appears in no relation?
+    bool found;
+    for (size_t gen = 0; gen < _nrgens; gen++) {
+      found = false;
+      for (const relation_t& rel : _relations) {
+        if (std::find(rel.first.cbegin(), rel.first.cend(), gen)
+            != rel.first.cend()) {
```
james-d-mitchell: You could combine this line with the next `if`:

    if (std::find(rel.first.cbegin(), rel.first.cend(), gen)
             != rel.first.cend() || std::find(rel.first.cbegin(), rel.first.cend(), gen) != rel.second.cend()) {

james-d-mitchell: Probably not worth it

PR #40, opened 2017-08-14
build: use rpmalloc instead of malloc (WIP)

james-d-mitchell:


james-d-mitchell:
I think this is ready to go, but I want to merge the benchmarking stuff in first, to double check that this is actually worth it.

james-d-mitchell:
This turns out to be slower than using the system malloc so, I'm going to close this PR.

PR #45, opened 2017-08-22
Add --enable-debug configure option

james-d-mitchell:


james-d-mitchell:
Gonna try something else

PR #48, opened 2017-09-04
WIP Add iterators to RecVec class

james-d-mitchell:
This PR adds iterators for the `Semigroup` and `RecVec` class, removes some superfluous member functions and data members and implements improvements in `sorted_elements` and the idempotent finding.


james-d-mitchell:
I expect the tests for GAP Semigroups package to fail here, since these changes remove some methods.

PR #49, opened 2017-09-06
Travis: add clang-format tests

mtorpey:
At long last, it works!

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -106,6 +106,7 @@ check-all: check
 
 format: 
 	@echo "Running clang-format . . .";  
+	@clang-format --version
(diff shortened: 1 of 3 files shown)


Makefile.am
```diff
@@ -106,6 +106,7 @@ check-all: check
 
 format: 
 	@echo "Running clang-format . . .";  
+	@clang-format --version
```
james-d-mitchell: Is this supposed to be here?

mtorpey:
That's intentional.  Given how much confusion was caused over what version of clang-format is running, I thought a final confirmation would be worth having.

PR #52, opened 2017-09-20
Permutation class

le27:
Adds the Permutation Class

### DIFF
--- a/src/elements.h
+++ b/src/elements.h
@@ -1251,6 +1251,36 @@ namespace libsemigroups {
     }
   };
 
+  //! Template class for permutations.
+  //!
+  //! The value of the template parameter \p T can be used to reduce the amount
+  //! of memory required by instances of this class; see PartialTransformation
+  //! and ElementWithVectorData for more details.
+  //!
+  //! A *permutation* \f$f\f$ is an injective transformation defined on the
+  //! whole of \f$\{0, 1, \ldots, n - 1\}\f$ for some integer \f$n\f$ called the
+  //! *degree* of \f$f\f$.
+  //! A permutation is stored as a vector of the images of
+  //! \f$(0, 1, \ldots, n - 1)\f$,
+  //! i.e. \f$((0)f, (1)f, \ldots, (n - 1)f)\f$.
+  template <typename T>
+  class Permutation : public Transformation<T> {
+   public:
+    using Transformation<T>::Transformation;
+    //! Returns the inverse of a permutation.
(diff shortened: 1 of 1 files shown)


src/elements.h
```diff
@@ -393,8 +393,14 @@ namespace libsemigroups {
     //! replaces the underlying vector of \c this with the underlying vector of
     //! \p x. Any method overriding this one must call
     //! Element::reset_hash_value.
+    //!
+    // TODO update the doc to reflect that this changes the degree of this if
+    // the degree of x is higher.
     void copy(Element const* x) override {
       LIBSEMIGROUPS_ASSERT(x->degree() == this->degree());
```
james-d-mitchell: This assertion directly contradicts the next line, so this will have one behaviour (assertion failure) when debugging is enabled, and another one (increasing the size) when it is not. There must not be any tests for this, because otherwise the CI would not have passed. I suggest deleting this change completely.

src/elements.h
```diff
@@ -567,6 +573,17 @@ namespace libsemigroups {
       return new T(vector);
     }
 
+    //! Returns boolean representing whether or not a partial transformation is
+    //! the identity.
+    bool is_identity() {
```
james-d-mitchell: It would be better if this was a deleted method for the `Element` class and there was a method for every type of `Element`. Again this does not seems related to the introduction of the `Permutation` class, so you might consider removing this completely from here, and making a separate PR with this complete change.

src/elements.h
```diff
@@ -823,6 +840,54 @@ namespace libsemigroups {
...
+  //! The value of the template parameter \p T can be used to reduce the amount
+  //! of memory required by instances of this class; see PartialTransformation
+  //! and ElementWithVectorData for more details.
+  //!
+  //! A *permutation* \f$f\f$ is an injective transformation defined on the
+  //! whole of \f$\{0, 1, \ldots, n - 1\}\f$ for some integer \f$n\f$ called the
+  //! *degree* of \f$f\f$.
+  //! A permutation is stored as a vector of the images of
+  //! \f$\{0, 1, \ldots, n - 1\}\f$,
+  //! i.e. \f$\{(0)f, (1)f, \ldots, (n - 1)f\}\f$.
```
james-d-mitchell: I'd use round brackets here not curly ones, since the order of these images matters.

james-d-mitchell:
Looks good please just add that one.blank line and I'll merge

PR #53, opened 2017-09-22
Refactor elements

james-d-mitchell:
This PR implements a substantial refactoring of the `Element` class and some of its subclasses, plus a couple of other minor changes. Highlights include:

This commit refactors the Element class and its subclass:

- the `swap` method is introduced, for swapping the contents of one
element with that of another. This method is used in
`Semigroup::word_to_element`

- constructors taking a hash value are removed, since these invalidate
the object if the hash value is incorrect, and so these were rather
dangerous.

- duplicate code was removed by using "using" to inherit constructors
from parent classes in several of the subclasses of `Element`

- `BooleanMat` is now a subclass of `MatrixOverSemiring`, and associated
changes, such as defining the `BooleanSemiring`, and changes to `Semiring`
and its derived classes to allow them to contain other types of
elements than `int64_t`. This removes a fair amount of duplicate code.

- duplicate code for hash functions was removed and some improvements
were made to the performance of hashing partial perms.


james-d-mitchell:
I expect the GAP tests to fail since this PR has changes that are not backwards compatible.

PR #70, opened 2018-01-16
bmat: add faster boolean matrices

james-d-mitchell:
This PR adds the class `BMat8` which provides faster boolean matrices.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -27,21 +27,20 @@ ACLOCAL_AMFLAGS	= -I m4
 ## libsemigroups headers
 pkginclude_HEADERS =  src/libsemigroups-config.h 
 pkginclude_HEADERS += src/libsemigroups-debug.h 
-pkginclude_HEADERS += src/blocks.h               src/cong.h 
-pkginclude_HEADERS += src/elements.h             src/semigroups.h 
-pkginclude_HEADERS += src/rws.h                  src/rwse.h 
-pkginclude_HEADERS += src/semiring.h             src/partition.h
-pkginclude_HEADERS += src/recvec.h               src/report.h	
-pkginclude_HEADERS += src/timer.h                src/uf.h
+pkginclude_HEADERS += src/blocks.h     src/bmat.h      src/cong.h
+pkginclude_HEADERS += src/elements.h   src/partition.h src/recvec.h
+pkginclude_HEADERS += src/report.h     src/rws.h       src/rwse.h
+pkginclude_HEADERS += src/semigroups.h src/semiring.h  src/timer.h 
+pkginclude_HEADERS += src/uf.h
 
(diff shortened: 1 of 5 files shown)


src/bmat.cc
```diff
@@ -0,0 +1,105 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+
+// This file contains an implementation of fast boolean matrices up to
+// dimension 8 x 8.
+
+#include "bmat.h"
+
+#include <algorithm>
```
flsmith: I don't think this include is needed

tests/bmat.test.cc
```diff
@@ -0,0 +1,151 @@
...
+             {1, 1, 1, 1, 1, 1, 1, 1},
+             {1, 1, 1, 1, 1, 1, 1, 1},
+             {1, 1, 1, 1, 1, 1, 1, 1},
+             {1, 1, 0, 1, 1, 1, 1, 1},
+             {0, 1, 1, 1, 0, 1, 1, 1},
+             {0, 1, 1, 1, 0, 1, 1, 1},
+             {1, 1, 1, 1, 1, 1, 1, 1}});
+
+  REQUIRE(tmp == bm3);
+  REQUIRE(tmp == bm * bm2);
```
flsmith: Should probably have more tests for multiplication, but I can't see a decent / non-trivial way of testing a lot of multiplications directly. The row space basis test did include a large number of multiplications.

Makefile.am
```diff
@@ -76,15 +76,15 @@ CHECK_FLAGS2  = [quick],[standard]
 
 check_PROGRAMS = lstest
 
-lstest_SOURCES =  tests/main.test.cc
+lstest_SOURCES =  tests/main.test.cc      tests/bmat.test.cc
```
flsmith: This isn't in the correct place. Do you care?

james-d-mitchell: Yes, probably

james-d-mitchell: Oh wait you mean `main.test.cc` is in the wrong place, I think we should leave it there

src/bmat.cc
james-d-mitchell: Will remove

tests/bmat.test.cc
james-d-mitchell: Probably but I am in the process of adding tests for in the `template` branch that will perform millions of multiplications, so maybe don't worry about this for now.

Makefile.am
flsmith: I agree with leaving the ``main.test.cc`` where it is, but then the ``bmat.test.cc`` is still placed inconsistently relative to the others

james-d-mitchell:
Superceded by #71

PR #71, opened 2018-01-16
bmat: add faster boolean matrices

flsmith:
Like [PR#70](https://github.com/james-d-mitchell/libsemigroups/pull/70), but with doc and minor fixes.


james-d-mitchell:
@flsmith travis seems to be having some trouble, but this failed on linting, can you please resolve this?

src/bmat.cc
```diff
@@ -0,0 +1,104 @@
...
+                                                 0x4040404040404040,
+                                                 0x2020202020202020,
+                                                 0x1010101010101010,
+                                                 0x808080808080808,
+                                                 0x404040404040404,
+                                                 0x202020202020202,
+                                                 0x101010101010101};
+
+  // BMat methods
+  //
```
james-d-mitchell: Can you delete these two comment lines please? They don't add much...

src/bmat.h
```diff
@@ -0,0 +1,222 @@
...
+      y = (x ^ (x >> 28)) & 0xF0F0F0F0;
+      x = x ^ y ^ (y << 28);
+      return BMat8(x);
+    }
+
+    //! Returns the matrix product of \c this and \p that
+    //!
+    //! This method returns the standard matrix product (over the
+    //! boolean semiring) of two BMat8.
+    // https://stackoverflow.com/a/18448513
```
james-d-mitchell: Might as well include the link in the actual doc, i.e. "From ..."

src/bmat.h
```diff
@@ -0,0 +1,222 @@
...
+      x = x ^ y ^ (y << 14);
+      y = (x ^ (x >> 28)) & 0xF0F0F0F0;
+      x = x ^ y ^ (y << 28);
+      return BMat8(x);
+    }
+
+    //! Returns the matrix product of \c this and \p that
+    //!
+    //! This method returns the standard matrix product (over the
+    //! boolean semiring) of two BMat8.
```
james-d-mitchell: "of two BMat8 instances/objects."

src/bmat.h
```diff
@@ -0,0 +1,222 @@
...
+    //!
+    //! Returns an unsigned integer obtained by interpreting an 8x8
+    //! BMat8 as a sequence of 64 bits (reading rows left to right,
+    //! from top to bottom)and then this sequence as an unsigned int.
+    inline uint64_t to_int() const { return _data; }
+
+    //! Returns the transpose of \c this
+    //!
+    //! Returns the standard matrix transpose of a BMat8.
+    inline BMat8 transpose() const {
```
james-d-mitchell: Reference for this?

src/bmat.h
```diff
@@ -0,0 +1,222 @@
...
+    friend std::ostream &operator<<(std::ostream &os, BMat8 const &bm) {
+      os << bm.to_string();
+      return os;
+    }
+
+    //! Returns a string representation of \c this
+    //!
+    //! This method returns a string which represents a BMat8, which may
+    //! for example be used to display the BMat8.
+    std::string to_string() const;
```
james-d-mitchell: I think we should replace this with a method for the `operator<<` of a `std::ostringstream`, and a `BMat8`.

src/bmat.h
```diff
@@ -0,0 +1,222 @@
...
+
+namespace libsemigroups {
+
+  //! Class for fast boolean matrices of dimension up to 8x8
+  //!
+  //! The methods for these small matrices over the boolean semiring
+  //! are more optimised than the generic methods for boolean matrices.
+  //! Note that all BMat8 are represented internally as an 8x8 matrix;
+  //! any entries not defined by the user are taken to be 0. This does
+  //! not affect the results of any calculations.
```
james-d-mitchell: Can we add that this is a trivial class?

src/bmat.h
```diff
@@ -0,0 +1,222 @@
...
+#include <functional>
+#include <iostream>
+#include <random>
+
+#include "libsemigroups-debug.h"
+#include "timer.h"
+
+namespace libsemigroups {
+
+  //! Class for fast boolean matrices of dimension up to 8x8
```
james-d-mitchell: `8x8` -> `8 x 8`

Makefile.am
```diff
@@ -27,21 +27,20 @@ ACLOCAL_AMFLAGS	= -I m4
 ## libsemigroups headers
```
james-d-mitchell: Can you change the entries in `Makefile.am` so that there is only one file per line, and they are in alphabetical order after `libsemigroups-config.h` and `libsemigroups-debug.h`? It is too large a pain to keep them in order otherwise :)

flsmith:
I've made the changes that you requested. If they look all right, I'll squash these commits as well.

As far as I can tell they do work in 32-bit - at least the tests pass in 32-bit mode on my machine. I haven't been able to compile the benchmarks in 32-bit mode though, so I have no idea how much performance might be affected. It's guess it's also possible that they only work in 32-bit mode on particular architectures.

src/bmat.h
```diff
@@ -0,0 +1,250 @@
...
+  class BMat8 {
+   public:
+    //! A default constructor.
+    //!
+    //! This constructor gives no guarantees on what the matrix will contain.
+    BMat8() = default;
+
+    //! A constructor.
+    //!
+    //! The 8 8-bit chunks of the binary representation of \p mat
```
james-d-mitchell: "8 chunks of 8-bits each"?

src/bmat.h
```diff
@@ -0,0 +1,250 @@
...
+      LIBSEMIGROUPS_ASSERT(0 <= i && i < 8);
+      LIBSEMIGROUPS_ASSERT(0 <= j && j < 8);
+      return (_data << (8 * i + j)) >> 63;
+    }
+
+    //! Returns the integer representation of \c this.
+    //!
+    //! Returns an unsigned integer obtained by interpreting an 8 x 8
+    //! BMat8 as a sequence of 64 bits (reading rows left to right,
+    //! from top to bottom)and then this sequence as an unsigned int.
```
james-d-mitchell: missing space

src/bmat.h
```diff
@@ -0,0 +1,250 @@
...
+        }
+        os << "\n";
+      }
+      return os;
+    }
+
+    //! Insertion operator
+    //!
+    //! This method allows BMat8 objects to be inserted into a ostream.
+    friend std::ostream &operator<<(std::ostream &os, BMat8 const &bm) {
```
james-d-mitchell: Is this method really required? If so, then I think you can change it back to 

~~~~
    friend std::ostream& operator<<(std::ostream& os, BMat8 const& bm) {
      os << bm.to_string();
      return os;
    }
~~~~

Since I think `to_string` just uses the previous `<<` method by default.

PR #72, opened 2018-01-16
Renner monoid of type B and D (EG and Gay-Hivert presentation)

hivert:
Dear James,

Thanks to you, we've been able to check everything we wanted ! It even allowed to detect and fix a small index mistake (a interval starting from 1 instead of 0) in the paper ! As you asked, I'm submitted a pull request but you don't yet should merge it (indeed travis will complain).

Some of the test seems to define infinite semigroups, so that Froidure-Pin will loop forever exhausting all the memory. Those tests (4 of them) are marked with a comment `// Loops for ever: Infinite monoid ???`. They should be fixed with something saying that there is at least a large number of elements. The current number marked as expected is the number which is wrongly claimed by E.G. So to prove what we need, say that the size is at least this number + 1 is ok.

Also I commented out two more tests which require 2 mins on my computer.

Again thanks a lot !

Best

Florent

### DIFF
--- a/tests/cong.test.cc
+++ b/tests/cong.test.cc
@@ -987,21 +987,6 @@ std::vector<size_t> max_elt_B(size_t i) {
   return t;
 }
 
-std::vector<size_t> max_elt_D(size_t i, int g) {
-// g est 0 ou 1 : 0 pour f et 1 pour e
-  std::vector<size_t> t{0};
-  int parity = g;
-  for (int end=i; end > 0; end--) {
-    t.push_back(parity);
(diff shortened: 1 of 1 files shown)


james-d-mitchell:
Great @hivert, many thanks!

hivert:
Added some more tests for two stages quotient !

james-d-mitchell:
While we didn’t merge this PR, the content of the PR will be included in the next release.

flsmith:
Closed as this is included in #116

PR #89, opened 2018-05-02
Replace really_delete from Element with shared_ptr

hivert:


james-d-mitchell:
This works but we (@hivert and I ) decided not to go down this route, since it appears that there is a performance penalty for using `shared_ptr` which is not acceptable.

PR #100, opened 2018-08-14
Schreier sims

aal20:
Adds `sims.h`, which includes the new `StabChain` class.

Includes the `group_size` method for calculating a permutation group's
size using the `Schreier-Sims` algorithm.

Also includes the `group_mem` method to the `StabChain` class, which uses the
`Schreier-Sims` algorithm to determine if a given permutation is in a
given permutation group.

A number of private functions are also included.

Includes unittests in `sims.test.cc`.

Written with @le27.


flsmith:
Closed as this is included in https://github.com/libsemigroups/libsemigroups/pull/116

PR #106, opened 2019-03-03
Provide pkg-config data and install it

dimpase:
With this PR:

~~~~
./autogen.sh && ./configure --prefix=foo && make install
~~~~

provide, on a (virtual) system with `pkg-config` knowing about "foo" goodies such as

$ pkg-config --modversion libsemigroups
1.0.0

(if "--prefix=" is omitted then, with sudo blessing, "make install" will land
/usr/local/lib/pkgconfig/libsemigroups.pc)

Also, autoconfing of libsemigroups as a dependency will be simplified
by calling PKG_CHECK_MODULES()

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,6 @@ gh-pages
 config.status
 config.log
 *~
-Makefile.in
 README
 aclocal.m4
 autom4te.cache/
@@ -26,6 +25,8 @@ configure
 *.la
 libtool
 *.in
+!libsemigroups.pc.in
+libsemigroups.pc
 autom4te.cache
 config
 .!*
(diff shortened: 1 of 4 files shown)


james-d-mitchell:
Thanks for the PR @dimpase can you please rebase this onto the `stable-0.6` branch? Thanks.

dimpase:
How about another PR against the stable branch- this seems to be the usual back-porting procedure?

james-d-mitchell:
Yes, that's ok, thanks.

dimpase:
done, see #108

james-d-mitchell:
I've cherry-picked the changes in this PR into the `master` branch, so I'll close this PR now.

PR #120, opened 2019-11-13
Enable code coverage

flsmith:
Will this be enough? Who knows.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,6 @@ services:
 
 matrix:
   include:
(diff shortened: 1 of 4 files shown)


james-d-mitchell:
Yeah I think I tried something like this before, and it didn't work :( I think it needs to be set up not as a travis job.

flsmith:
I think this is ready.

james-d-mitchell:
Thanks @flsmith

PR #121, opened 2019-11-13
Added binder configuration, badge, and demo notebook

nthiery:
Untested with libsemigroups 1.0


james-d-mitchell:
Thanks @nthiery, I'll merge and see what happens!

james-d-mitchell:
@nthiery it works! But because we haven't actually released v1.0.0 yet the conda package is still using v0.6.7, but in any case it works which is awesome.

nthiery:
> @nthiery it works! But because we haven't actually released v1.0.0
> yet the conda package is still using v0.6.7, but in any case it
> works which is awesome.

It's pretty cool technology indeed :-)

PR #138, opened 2019-12-05
Fix test failures on 32-bit systems

jamesjer:
The first commit fixes issue 60.  It appears that when the first two iterators passed to std::equal are in reverse order, std::equal sometimes returns true and sometimes returns false.  I have only seen it return true on 32-bit x86 systems, and this may very well be a bug in the version of the C++ library I am using.  In any case, this commit seems to me like the right thing to do, and it resolves the issue.

The second commit fixes test failures due to integer overflow.  Note that size_t is a 32-bit type on 32-bit Linux systems.

### DIFF
--- a/include/schreier-sims.hpp
+++ b/include/schreier-sims.hpp
@@ -368,16 +368,16 @@ namespace libsemigroups {
 
     //! Returns the size of the group represented by this.
     //!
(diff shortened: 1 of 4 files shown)


james-d-mitchell:
Thanks @jamesjer this is awesome!

PR #147, opened 2020-01-09
External fmt

james-d-mitchell:
I think, this PR fully resolves Issue #145. I need to add a CI job to check that this config option works.


james-d-mitchell:
@isuruf I think this is ready to go, can you please confirm that this fully resolves Issue #145 ? Thanks!

PR #148, opened 2020-01-10
Attempt to increase abi version number

james-d-mitchell:
This is an attempt to resolve Issue #144.


james-d-mitchell:
@isuruf I think this resolves Issue #144, can you confirm please?

PR #170, opened 2020-01-16
Prefix all includes with libsemigroups/

nthiery:
Then there is no need to add libsemigroups/ to the include path for a software (e.g. libsemigroups_cppyy) to include libsemigroups's headers.


nthiery:
Superseded by https://github.com/libsemigroups/libsemigroups/pull/172

PR #172, opened 2020-01-18
Includes in libsemigroups/

nthiery:
Goal: encourage libsemigroups users to include them as

    #include "libsemigroups/..."

- Move includes from includes/*.hpp to includes/libsemigroups/*.hpp
- Update source and test files to use the above include directive


nthiery:
The single failure is with the gap travis test. My interpretation is that it's just that the configure.ac of gap's libsemigroup module needs to be updated to point to libsemigroups/libsemigroups.hpp.

james-d-mitchell:
Thanks for this @nthiery, the CI fails, but this is expected since the version of Semigroups in GAP used in the CI is not compatible with this change. I'll make some comments in a moment about how to go about remedying this.

james-d-mitchell:
You're probably right @nthiery, I'd like to see the CI passing before merging, for this it should be sufficient to push some changes to GAP Semigroups to a branch, and then point the CI scripts to that branch rather than the release branch of Semigroups.

nthiery:
> You're probably right @nthiery, I'd like to see the CI passing
> before merging, for this it should be sufficient to push some
> changes to GAP Semigroups to a branch, and then point the CI scripts
> to that branch rather than the release branch of Semigroups.

Ok! Do I need to take action, or will you handle it?

james-d-mitchell:
@nthiery I can make the changes to Semigroups in GAP but you'll need to update this PR to use the appropriate branch/repo for Semigroups in the CI.

james-d-mitchell:
I'm happy to merge this, but can you drop the temporary fix to the CI (I've merged the PR to Semigroups in GAP)? Thanks @nthiery

nthiery:
Ok! I forced pushed the branch with the last commit containing the temporary fix to CI popped out. Should be good to go.

Thanks!

PR #175, opened 2020-01-28
travis: add tests for libsemigroups_cppyy

james-d-mitchell:


james-d-mitchell:
The travis "gap" test suite currently fails, but will pass when:

Issue #171, opened 2020-01-17
Iterate through words in shortlex order


hivert:

james-d-mitchell:
Resolved by v1.1.0

Issue #169, opened 2020-01-16
Rename shortlex_words to words_in_shortlex_order


hivert:

james-d-mitchell:
Resolved by v1.1.0

PR #195, opened 2020-03-27
Add functionality for iterating through words & string in lex or short-lex order

james-d-mitchell:


james-d-mitchell:
Resolves #171 and #169

PR #208, opened 2020-05-27
Improve IsObviouslyInfinite

reiniscirpons:
This PR adds additional checks to `IsObviouslyInfinite` and adds tests for it.

`IsObviouslyInfinite` now checks whether there are more relations than generators, as well as adding a new matrix based test and a check for free products. In addition a new test file for `IsObviouslyInfinite` was added. As a result of the changes, the Bicyclic monoid now is correctly identified as obviously infinite and test case `Congruence 014` was changed to reflect this.

This PR also introduces a dependency on [Eigen](http://eigen.tuxfamily.org/) for handling matrices and rank calculations. The build system has not been configured yet so I expect the external checks to fail.


james-d-mitchell:
Superseded by the other pr.

PR #218, opened 2020-06-23
Add Gilman automaton and associated functionality

james-d-mitchell:
This is essentially done, I have one FIXME to fix, and want to see how the CI runs, then do some rebasing.


saraedum:
@slel made me aware of this PR. I don't know what you are doing really but I could comment on cereal here since I had long fights with it over at libflatsurf and its dependencies.

saraedum:
Please let me know if you want me to have a look. Or if you prefer a more interactive format, I'm also at https://zulip.sagemath.org/#narrow/stream/119-polygon/topic/Serialization.

james-d-mitchell:
@saraedum and @slel thanks so much for the comments! I'd really value that conversation @saraedum, I didn't see your comments in time. For the time being I think I'm going to drop cereal from this PR, hopefully we can have a discussion, then I can think about opening a new PR with adding cereal (or something else).

james-d-mitchell:
This PR is superseded by PR #219

saraedum:
> @saraedum and @slel thanks so much for the comments! I'd really value that conversation @saraedum, I didn't see your comments in time. For the time being I think I'm going to drop cereal from this PR, hopefully we can have a discussion, then I can think about opening a new PR with adding cereal (or something else).

We can talk about this in more detail but here are some things I found through a lot of trial and error:

I ship cereal headers as part of the tests, e.g., [here](https://github.com/flatsurf/flatsurf/blob/master/libflatsurf/test/Makefile.am#L26), but do not vendor them in any way when installing C++ libraries. For one, it's annoying to have them in your path in `include/`. But more importantly, some versions are not fully compatible so you want your users to provide cereal headers that work for them. 

As a consequence, everything cereal needs to be in a separate header of the project that is included to opt into cereal support. The nice thing is that this means that you don't need to detect cereal at build time anymore :) [Here](https://github.com/flatsurf/flatsurf/blob/master/libflatsurf/flatsurf/cereal.hpp) is an example of such a standalone header with some hacks to work around problems in cereal…

This is a problem when cereal is needed in a compiled library, e.g., to serialize internal types. I have not found a great solution for that, though I have some ideas. [Here](https://github.com/flatsurf/exact-real/issues/89) is an example of this.

If your C++ code supports cereal, you can use it to get Python pickling with some caveats mostly for free with [cppyythonizations](https://github.com/flatsurf/cppyythonizations/blob/master/src/cppyythonizations/pickling/cereal.py).

PR #233, opened 2020-08-27
Config options exit 1

james-d-mitchell:


james-d-mitchell:
Just testing that travis shows a failure when there is one.

PR #237, opened 2020-08-31
Set eigen_enabled to true if --with-external-eigen is given.

jamesjer:
Currently, passing --with-external-eigen to configure results in eigen support being disabled.


james-d-mitchell:
Thanks so much @jamesjer, I think the CI errors are spurious, I'm happy to merge this.

PR #242, opened 2020-09-30
Fix a few bugs in PBRs

flsmith:
1) Display was broken for certain PBRs
2) PBRs weren't required to have sorted adjacencies, which could result in mathematically equal PBRs being nonequal.


src/element.cpp
```diff
@@ -649,6 +659,14 @@ namespace libsemigroups {
         }
       }
     }
+    for (size_t u = 0; u < n; ++u) {
+      if (!std::is_sorted(this->_vector.at(u).cbegin(),
+                          this->_vector.at(u).cend())) {
+        LIBSEMIGROUPS_EXCEPTION("the adjacencies of vertex ",
+                                detail::to_string(u),
```
james-d-mitchell, line 666: This needs to be `detail::to_string(u).c_str()` to work when `fmt` is not enabled.

james-d-mitchell:
```suggestion
                                detail::to_string(u).c_str(),
```

src/element.cpp
```diff
@@ -860,13 +878,17 @@ namespace libsemigroups {
...
-        if (x < 0) {
-          v.push_back(static_cast<uint32_t>(n - x - 1));
-        }
         if (x > 0) {
           v.push_back(static_cast<uint32_t>(x - 1));
         }
       }
+      // We have to go backwards through the vector to add the negative entries,
+      // so that users can input those negative entries in the natural order
+      for (auto it = vec.rbegin(); it < vec.rend(); ++it ) {
```
flsmith, line 887:
```suggestion
      for (auto it = vec.rbegin(); it < vec.rend(); ++it) {
```

src/element.cpp
```diff
@@ -881,13 +903,15 @@ namespace libsemigroups {
...
               x);
         }
-        if (x < 0) {
-          v.push_back(static_cast<uint32_t>(n - x - 1));
-        }
         if (x > 0) {
           v.push_back(static_cast<uint32_t>(x - 1));
         }
       }
+      for (auto it = vec.rbegin(); it < vec.rend(); ++it ) {
```
flsmith, line 910:
```suggestion
      for (auto it = vec.rbegin(); it < vec.rend(); ++it) {
```

PR #260, opened 2020-12-08
Better matrix mult

james-d-mitchell:


james-d-mitchell:
I think this is good to go.

PR #261, opened 2020-12-23
Delete uf old

james-d-mitchell:
Move the files `uf-old.hpp` and `uf-old.cpp` into the benchmarks, and also move the old benchmarks into the benchmarks directory.


james-d-mitchell:
This will fail because Semigroups for GAP is not yet updated.

PR #268, opened 2021-02-04
Include ./autogen.sh in the distro

james-d-mitchell:
This PR adds the `./autogen.sh` file to the distro. 

Rationale: the `make distclean` target deletes files created by `./autogen.sh` (namely `configure` and `.VERSION`) and is always present in the distro for `libsemigroups`, so without the `autogen.sh` file it is always possible to run `make distclean` in a release version of `libsemigroups` and not have a mechanism to regenerate these files.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -958,7 +958,6 @@ clean-local-check:
 distclean-local:
 	rm -f configure
 	rm -f include/libsemigroups/libsemigroups-config.hpp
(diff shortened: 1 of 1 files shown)


james-d-mitchell:
This PR should resolve:

https://github.com/semigroups/Semigroups/issues/740 

(once a new version of both libsemigroups and Semigroups are released).

PR #279, opened 2021-03-22
Check for C headers, types, and functions in C mode

jamesjer:
I noticed that some configure checks were failing, e.g.:
```
checking for stdbool.h that conforms to C99... no
checking for _Bool... no
checking for inline... inline
checking for int64_t... yes
checking for size_t... yes
checking for uint64_t... yes
checking for ptrdiff_t... no
```

The problem is that `AC_LANG([C++])` shifts autoconf into C++ mode, so it performs those checks with g++.  The fix is to be in C mode while checking C headers, types, and functions, then shift back to C++ mode.  With this change, those configure tests pass.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -25,6 +25,7 @@ AC_LANG([C++])
 AX_CXX_COMPILE_STDCXX_14(,[mandatory])
 
 dnl ## Check for pthread, this seems to be required to compile with GCC
+AC_LANG_PUSH([C])
 AX_PTHREAD(,[AC_MSG_ERROR([pthread is required])])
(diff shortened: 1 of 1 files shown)


james-d-mitchell:
@jamesjer thanks so much for the PR, this is most welcome. I completely failed to notice that this existed until just now, the changes look good to me, but I'm not going to merge this until I manage to fix the CI (the failures in it don't seem related to this PR at all).

jamesjer:
No problem.  I overlook things all the time. :-)

james-d-mitchell:
Thanks again @jamesjer, most welcome!

PR #280, opened 2021-04-19
Fix coverage target

james-d-mitchell:


james-d-mitchell:
I seem to have fixed this in #270 so I'm going to close this, in favour of #270.

PR #283, opened 2021-05-22
doc: try to use doxygen 1.9.1 on rtd

james-d-mitchell:


james-d-mitchell:
I'm going to close this temporarily until I fix the issues with RTD so as to avoid running and re-running the CI.

PR #297, opened 2021-06-09
konieczny: store lambda/rho positions for D class reps

flsmith:
Improves performance by reducing superfluous computation of lambda/rho values for D class representatives.

### DIFF
--- a/include/libsemigroups/konieczny.hpp
+++ b/include/libsemigroups/konieczny.hpp
@@ -237,6 +237,10 @@ namespace libsemigroups {
     using lambda_orb_scc_index_type = typename lambda_orb_type::scc_index_type;
     using rho_orb_index_type        = typename rho_orb_type::index_type;
     using rho_orb_scc_index_type    = typename rho_orb_type::scc_index_type;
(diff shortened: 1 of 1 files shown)


include/libsemigroups/konieczny.hpp
```diff
@@ -237,6 +237,10 @@ namespace libsemigroups {
     using lambda_orb_scc_index_type = typename lambda_orb_type::scc_index_type;
     using rho_orb_index_type        = typename rho_orb_type::index_type;
     using rho_orb_scc_index_type    = typename rho_orb_type::scc_index_type;
+    using rep_info_type             = typename std::tuple<internal_element_type,
```
james-d-mitchell, line 240: It probably doesn't make much odds, but could you please use an inner struct instead of a tuple here?

james-d-mitchell (changes requested):
Thanks @flsmith looks like a good improvement, be good to have an indication somewhere of what difference this PR makes. The only change I want is to not use a `tuple` but rather a nested `struct` with meaningful data member names.

PR #298, opened 2021-06-30
Implement EqualInFreeband

reiniscirpons:
This pull request is joint with @james-d-mitchell, @MTWhyte and @tomcontileslie. It implements a linear time method for checking equality of words as elements of the free band from [Radoszewski, Rytter (2010)](https://link.springer.com/chapter/10.1007/978-3-642-11266-9_55), see linked article for more info.

To check that it truly is linear, we ran a number of benchmarks on random words over a given alphabet and of a given length.
Figure_1
From Figure 1 it can be seen that the method seems to be linear in the length.
Figure_2
Figure 2 does not show a clear linear trend between alphabet size and runtime, but it seems that the runtime decreases as alphabet size increases. This is partially because each datapoint comes from a single run on a single word, and partially because random words are not the best test cases (as it is likely that a very small prefix of the word contains almost all letters of the word, for example).
Figure_3
Figure 3 is a 3d scatter plot  of the runtimes together with a least squares estimate of a best fit plane (generated with code taken from [here](https://gist.github.com/RustingSword/e22a11e1d391f2ab1f2c)), and visually looks like an ok fit. Maybe more tests should be run to determine if it truly is linear.

What is clear though is that this method runs far better than the naive method that is exponential in the size of the alphabet (and thus cannot handle the case of size 50 alphabet).


benchmarks/bench-freeband.cpp
```diff
@@ -0,0 +1,159 @@
+//
+// libsemigroups - C++ library for semigroups and monoids
+// Copyright (C) 2020 James D. Mitchell
```
james-d-mitchell, line 3:
```suggestion
// Copyright (C) 2021 James D. Mitchell
```

include/libsemigroups/freeband.hpp
```diff
@@ -0,0 +1,129 @@
+//
+// libsemigroups - C++ library for semigroups and monoids
+// Copyright (C) 2020 James D. Mitchell
```
james-d-mitchell, line 3:
```suggestion
// Copyright (C) 2021 James D. Mitchell
```

src/freeband.cpp
```diff
@@ -0,0 +1,271 @@
+//
+// libsemigroups - C++ library for semigroups and monoids
+// Copyright (C) 2020 James D. Mitchell
```
james-d-mitchell, line 3:
```suggestion
// Copyright (C) 2021 James D. Mitchell
```

tests/test-freeband.cpp
```diff
@@ -0,0 +1,177 @@
+//
+// libsemigroups - C++ library for semigroups and monoids
+// Copyright (C) 2020 James D. Mitchell
```
james-d-mitchell, line 3:
```suggestion
// Copyright (C) 2021 James D. Mitchell
```

PR #334, opened 2022-03-01
ci: add cygwin job

james-d-mitchell:


wilfwilson:
The error that you're getting looks like something to do with Windows versus Unix line endings.

I recommend adding the same `.gitattributes` file to the root directory of this repository that I add in https://github.com/semigroups/Semigroups/pull/787. This ensures (I think?) that all files are checked out with Unix line endings. In any case, it might help.

.github/workflows/runtests.yml
```diff
@@ -2,10 +2,42 @@ name: CI
 on: [pull_request]
 
 jobs:
+   cygwin:
+      timeout-minutes: 40
+      runs-on: windows-latest
```
wilfwilson, line 7: I might not have got the syntax right, but I thin that adding this would stop you from having to manually configure the `shell` on each step.
```suggestion
      runs-on: windows-latest
      defaults:
        run:
          shell: C:\cygwin64\bin\bash.exe --login -o igncr '{0}'
```

james-d-mitchell:
Thanks @wilfwilson that's awesome, I'll have another stab later!

james-d-mitchell:
The second commit in this PR also resolves #335

PR #348, opened 2022-05-18
Add static triangular array container

Joseph-Edwards:
We sometimes use triangular matrices, but implementing them using the classic `StaticVector2` allocates more memory than is ever used. We implement `StaticTriArray2` to provide a memory efficient alternative, using the same interface as `StaticVector2`.

### DIFF
--- a/include/libsemigroups/containers.hpp
+++ b/include/libsemigroups/containers.hpp
@@ -36,8 +36,88 @@
 
 namespace libsemigroups {
   namespace detail {
+    template <typename T, size_t N>
+    class StaticTriVector2 {
+      // So that StaticTriVector2<T, N> can access private data members
+      // of StaticTriVector2<S, M> and vice versa.
+      template <typename S, size_t M>
+      friend class StaticTriVector2;
 
-    // Template class for 2-dimnensional dynamic arrays.
+     public:
+      // Aliases for iterators
+      using const_iterator =
+          typename std::array<T, N*(N + 1) / 2>::const_iterator;
+
+      StaticTriVector2(StaticTriVector2 const&) = default;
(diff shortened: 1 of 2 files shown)


include/libsemigroups/containers.hpp
```diff
@@ -30,6 +30,88 @@
 
 namespace libsemigroups {
   namespace detail {
+    template <typename T, size_t N>
+    class StaticTriArray2 {
```
reiniscirpons, line 34:
```suggestion
    class StaticTriVector2 {
```
Change name to better reflect that the class implements a `Vector2` interface.

include/libsemigroups/containers.hpp
```diff
@@ -30,6 +30,88 @@
...
+      // So that StaticTriArray2<T, N> can access private data members
+      // of StaticTriArray2<S, M> and vice versa.
+      template <typename S, size_t M>
+      friend class StaticVector2;
+
```
reiniscirpons, line 39:
```suggestion
      template <typename S, size_t M>
      friend class StaticTriVector2;

```
Your probably want the firend class to have the same name as the current class, since `StaticVector2` has different private data members to your class.

reiniscirpons (commented):
Thanks for the PR Joe! Ive tried using it in Schreier-Sims and it works just fine. Ive suggested some minor reworks, but otherwise it looks fine to me.

It would also be nice to have the iterator tests, but im not sure if its critical, since `StaticVector2` doesnt implement them either.

include/libsemigroups/containers.hpp
Joseph-Edwards: Thanks for pointing this out! This name change has been made.

PR #350, opened 2022-06-08
Add permutation group intersections

reiniscirpons:
This PR introduces a Schreier-Sims helper class and implements a helper function for computing the intersection of two permutation groups given as Schreier-Sims objects. Follows the general backtrack method given in Chapter 9.1 of [^1].

A number of improvements are still possible:
1.  A problem specific heuristic which reduces the number of branches to consider by looking at the intersection of orbit points (sufficiently moved), as described in Example 3 of Chapter 9.1.2 of [^1].
2. The method actually proceeds by finding successively larger stabilizers of the intersection with respect to the same base as one of the original groups. Currently we do not use this and simple gather up all the generators of the intersection before running the Schreier-Sims algorithm. We could skip running the algorithm by building the Schreier-Sims object as we go along. Not sure how big of a time save this would be though.

Also, for some reason I cannot simultaneously rebase the master and the perm-inter branch so there might be minor headaches to whomever tries to merge this into master.

[^1]: Ákos Seress, "Permutation Group Algorithms" (2003)


james-d-mitchell:
Before reviewing this any further, here are a couple of comments:

1) Please rebase onto the most recent `master` branch, your PR at present includes lots of unrelated commits.

2) The `distcheck` CI job is probably failing because `schreier-sims-helper.hpp` is not listed as one of the includes for the package in `Makefile.am`. Please add it there and try again.

3) The `valgrind` CI job is failing because some of your tests take too long to run when using valgrind, for example,

~~~
SchreierSimsHelper 010: S17 and A39 intersection . . .5m51s
~~~

if you add the tag `[no-valgrind]` to any long running tests, then this CI job will skip them.

reiniscirpons:
Thanks for the helpful comments @james-d-mitchell . I added the include in `Makefile.am` and also added the `[no-valgrind]` tag to tests that took longer than 100ms. Hopefully this makes it so that the tests pass now.

I was however unable to rebase onto master. Or rather, I *can* rebase onto master but then I cannot push into `perm-inter` and need to do another rebase, which then means I need to rebase onto master again and so on in perpetuity. I spent a while trying to troubleshoot it (it might be related to that bugfix that you implemented for `add_point`), but figured it would just be easier to start on a new branch/PR. So see #351 for the second attempt. I will close this PR not to confuse anyone.

PR #351, opened 2022-06-09
Add permutation group intersections (for real this time)

reiniscirpons:
This PR introduces a Schreier-Sims helper class and implements a helper function for computing the intersection of two permutation groups given as Schreier-Sims objects. Follows the general backtrack method given in Chapter 9.1 of [^1].

A number of improvements are still possible:

A problem specific heuristic which reduces the number of branches to consider by looking at the intersection of orbit points (sufficiently moved), as described in Example 3 of Chapter 9.1.2 of [^1].
The method actually proceeds by finding successively larger stabilizers of the intersection with respect to the same base as one of the original groups. Currently we do not use this and simple gather up all the generators of the intersection before running the Schreier-Sims algorithm. We could skip running the algorithm by building the Schreier-Sims object as we go along. Not sure how big of a time save this would be though.

I also updated my personal webpage in the readme and docs, hope thats ok.

(This is a new PR on the same topic as #350 since I messed up rebasing before and it was easier to just make a new branch than salvage whatever was there)

[^1]: Ákos Seress, "Permutation Group Algorithms" (2003)


docs/source/api/schreiersims-helper.rst
```diff
@@ -0,0 +1,22 @@
+.. Copyright (c) 2020, J. D. Mitchell
```
james-d-mitchell, line 1:
```suggestion
.. Copyright (c) 2022, Reinis Cirpons
```

include/libsemigroups/schreier-sims-helper.hpp
```diff
@@ -0,0 +1,182 @@
+// libsemigroups - C++ library for semigroups and monoids
+// Copyright (C) 2021 Reinis Cirpons
```
james-d-mitchell, line 2:
```suggestion
// Copyright (C) 2022 Reinis Cirpons
```
Time is always

include/libsemigroups/schreier-sims-helper.hpp
```diff
@@ -0,0 +1,182 @@
...
+#define LIBSEMIGROUPS_SCHREIER_SIMS_HELPER_HPP_
+
+#include <cstddef>  // for size_t
+
+#include "containers.hpp"     // for Array2
+#include "schreier-sims.hpp"  // for SchreierSims, SchreierSims<>::ele...
+#include "transf.hpp"         // for LeastPerm
+#include "types.hpp"          // for SmallestInteger
+
+#include <iostream>
```
james-d-mitchell, line 37: Should this be included? If it should, then it should be before the `libsemigroups` headers.

include/libsemigroups/schreier-sims-helper.hpp
```diff
@@ -0,0 +1,182 @@
...
+
+namespace libsemigroups {
+
+  namespace schreier_sims_helper {
+
+    template <size_t N>
```
james-d-mitchell, line 43: Why not put these aliases inside the function `intersection`? This way they aren't exposed to the outside

include/libsemigroups/schreier-sims.hpp
```diff
@@ -325,16 +325,22 @@ namespace libsemigroups {
...
     //!
     //! \complexity
     //! Constant.
     //!
     //! \parameters
     //! (None)
-    // TODO(later) shouldn't this throw if depth is out of bounds??
     size_t number_of_strong_generators(index_type depth) const noexcept {
+      if (_base_size == 0)
+        return 0;
```
james-d-mitchell, line 337:
```suggestion
      if (_base_size == 0) {
        return 0;
      }
```

include/libsemigroups/schreier-sims.hpp
```diff
@@ -325,16 +325,22 @@ namespace libsemigroups {
...
+      if (depth >= _base_size) {
+        LIBSEMIGROUPS_EXCEPTION(
+            "depth out of bounds, expected value in range [0, %d), got %d",
+            _base_size,
```
james-d-mitchell, line 341:
```suggestion
            uint64_t(_base_size),
```

include/libsemigroups/schreier-sims.hpp
```diff
@@ -325,16 +325,22 @@ namespace libsemigroups {
...
+            depth);
```
james-d-mitchell, line 342:
```suggestion
            uint64_t(depth));
```

include/libsemigroups/schreier-sims.hpp
```diff
@@ -363,6 +375,88 @@ namespace libsemigroups {
...
+    //!
+    //! \returns
+    //! A boolean indicating if the point \p pt is in the orbit of the
+    //! basepoint of \c this at depth \p depth.
+    //!
+    //! \throws LibsemigroupsException if the \p depth is out of bounds.
+    //!
+    //! \complexity
+    //! Constant.
+    bool orbits_lookup(index_type depth, point_type pt) {
```
james-d-mitchell, line 450: I think you've added documentation here to things that weren't previously documented, right? So, please make sure they actually show up in the doc when it's compiled, you probably need to add some entries in the file `docs/yml/schreiersims.yml` to make that happen.

include/libsemigroups/schreier-sims.hpp
```diff
@@ -363,6 +375,88 @@ namespace libsemigroups {
...
+      if (depth >= _base_size) {
+        LIBSEMIGROUPS_EXCEPTION(
```
james-d-mitchell, line 452: Same comment about %d

james-d-mitchell: Also are there any tests for these lines?

include/libsemigroups/schreier-sims-helper.hpp
```diff
@@ -0,0 +1,182 @@
...
+
+    //! Find the intersection of two permutation groups.
+    //!
+    //! Modifies the first parameter \p T to be the Schreier-Sims object
+    //! corresponding to the intersection of \p S1 and \p S2.
+    //!
+    //! \tparam N the largest point not fixed by the permutations in the
+    //! permutation groups.
+    //!
+    //! \param T the Screier-Sims object holding the result.
```
james-d-mitchell, line 65:
```suggestion
    //! \param T the Schreier-Sims object holding the result.
```

include/libsemigroups/schreier-sims-helper.hpp
```diff
@@ -0,0 +1,182 @@
...
+
+      SchreierSims<N> S2B;
+      for (size_t depth = 0; depth < base_size; ++depth) {
+        S2B.add_base_point(S1.base(depth));
+      }
+      for (size_t i = 0; i < S2.number_of_generators(); ++i) {
+        S2B.add_generator(S2.generator(i));
+      }
+      S2B.run();
+
```
james-d-mitchell, line 102:
```suggestion
#ifdef LIBSEMIGROUPS_DEBUG
```

include/libsemigroups/schreier-sims-helper.hpp
```diff
@@ -0,0 +1,182 @@
...
+      for (size_t depth = 0; depth < base_size; ++depth) {
+        LIBSEMIGROUPS_ASSERT(S1.base(depth) == S2B.base(depth));
+      }
+
```
james-d-mitchell, line 106:
```suggestion
#endif
```

include/libsemigroups/schreier-sims-helper.hpp
```diff
@@ -0,0 +1,182 @@
...
+      if (N <= 1) {
+        T.run();
+        return;
+      }
+
+      // Note that if N-1 points are fixed then the N-th point is also fixed.
+      // So if base contains all N points, then we lose nothing by discarding
+      // the last point in the base.
+      size_t base_size = S1.base_size();
+      if (base_size == N)
```
james-d-mitchell, line 91: Missing brackets `{`

include/libsemigroups/schreier-sims-helper.hpp
```diff
@@ -0,0 +1,182 @@
...
+      // If N <= 1 then both S1, S2 are trivial.
+      if (N <= 1) {
+        T.run();
+        return;
+      }
+
+      // Note that if N-1 points are fixed then the N-th point is also fixed.
+      // So if base contains all N points, then we lose nothing by discarding
+      // the last point in the base.
+      size_t base_size = S1.base_size();
```
james-d-mitchell, line 90:
```suggestion
      size_t const base_size = S1.base_size();
```

include/libsemigroups/schreier-sims-helper.hpp
james-d-mitchell: Or you could delete this loop altogether, it looks like you check for the same thing lower down.

include/libsemigroups/schreier-sims-helper.hpp
```diff
@@ -0,0 +1,182 @@
...
+      // This is not currently implemented! We just use all of the points
+      // in the orbits of S1. Implementing it probably requires refactoring the
+      // code.
+      detail::StaticVector2<point_type<N>, N> refined_orbit;
+      for (size_t depth = 0; depth < base_size; ++depth) {
+        // First point is always base point to make algorithm simpler
+        LIBSEMIGROUPS_ASSERT(S1.base(depth) == S2B.base(depth));
+        refined_orbit.push_back(depth, S1.base(depth));
+        for (point_type<N> pt = 0; pt < N; ++pt) {
+          if ((pt != S1.base(depth)) && S1.orbits_lookup(depth, pt))
```
james-d-mitchell, line 121: Missing `{`

include/libsemigroups/schreier-sims-helper.hpp
```diff
@@ -0,0 +1,182 @@
...
+
+        // Find largest depth that has an unvisited node and increment its
+        // index. Adjust stabilizer depth as depths are exhausted.
+        for (;; --depth) {
+          LIBSEMIGROUPS_ASSERT((depth >= 0) && (depth < base_size));
+          state_index[depth]++;
+          if (state_index[depth] < refined_orbit.size(depth)) {
+            break;
+          }
+          if (depth < stab_depth)
```
james-d-mitchell, line 168:
```suggestion
          if (depth < stab_depth) {
```

include/libsemigroups/schreier-sims-helper.hpp
```diff
@@ -0,0 +1,182 @@
...
+            stab_depth = depth;
+          state_index[depth] = 0;
+          state_elem[depth]  = One<N>()(N);
+          if (depth == 0)
```
james-d-mitchell, line 172:
```suggestion
          if (depth == 0) {
```

include/libsemigroups/schreier-sims-helper.hpp
james-d-mitchell: Also as a separate comment: what pre-conditions exist for `T`? Does `T` have to be empty? If so, this should be written in the doc, and checked in the code.

include/libsemigroups/schreier-sims-helper.hpp
```diff
@@ -0,0 +1,182 @@
...
+      }
+
+      // Note that if N-1 points are fixed then the N-th point is also fixed.
+      // So if base contains all N points, then we lose nothing by discarding
+      // the last point in the base.
+      size_t base_size = S1.base_size();
+      if (base_size == N)
+        base_size = N - 1;
+
+      SchreierSims<N> S2B;
```
james-d-mitchell, line 94: If `N` is big enough, then this will overflow the stack. Maybe `S2B` should be allocated on the heap, `auto S2B = new SchreierSims<N>();` or even better use a `unique_ptr` so you don't have to remember to delete this later. Alternatively, you might consider having two variants of this, one where S2B is passed as a parameter, and another where it is not (i.e. like the current version but where you pass it your newly allocated object).

james-d-mitchell (changes requested):
Looks pretty good, a host of minor things to fix, then we're good to go!

include/libsemigroups/schreier-sims-helper.hpp
reiniscirpons: No, this was just from my poor attempts at debugging earlier (before being enlightened to the wonders of `gdb`). I've removed it now.

include/libsemigroups/schreier-sims-helper.hpp
reiniscirpons: Has to be empty. Added check to code and documentation.

include/libsemigroups/schreier-sims-helper.hpp
reiniscirpons: I did the heap allocation with the [`make_unique`](https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique) helper function.

include/libsemigroups/schreier-sims.hpp
reiniscirpons: Added tests

PR #359, opened 2022-08-31
Shrink to rows

james-d-mitchell:
Add a feature to `DynamicArray2` and another minor change to the files in the `etc` dir


flsmith (approved):
LGTM

Some file changes don't seem strongly related, but as long as there's no squash it's all good.

PR #365, opened 2022-09-07
Partially fix empty string segfault in `freeband_equal_to`.

reiniscirpons:
Theres a segfault which occurs in `freeband_equal_to` if one of the parameters is the empty string. This is maybe theoretically not an issue since the empty word is technically not a part of the free band if we think of it as a the free object in a variety of semigroups as opposed to monoids. On the other hand since we dont lose anything by including the empty word, I've modified the code to now work in the free band with identity adjoined.

This isn't a full fix though since we just add a check which filters the empty word, without removing a possible unsafe memory access in some other underlying function.

### DIFF
--- a/src/freeband.cpp
+++ b/src/freeband.cpp
@@ -238,6 +238,11 @@ namespace libsemigroups {
     if (x == y) {
       return true;
     }
+    if (x.empty() || y.empty()) {
+      // FIXME: Code segfaults without this check, its possible there is some
+      // unsafe memory access further in the code.
+      return false;
+    }
     size_t N = content_size(x);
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
Thanks @reiniscirpons the "partially" makes me think, maybe you intended to do some more stuff here? Is that the case, or can I merge?

reiniscirpons:
@james-d-mitchell you can merge. I would like to go over it with a more fine tooth comb some other time to figure out what exactly caused the segfault, but it seems like only the empty string corner case was causing issues so I think its not too important at the moment.

PR #366, opened 2022-09-11
Rc v2.2.2

james-d-mitchell:


james-d-mitchell:
The `libsemigroups_pybind11` test will fail unavoidably since the previous output (which the failing test contains) was incorrect.

PR #368, opened 2022-09-14
Improve `equal_in_freeband` and the coverage test script

reiniscirpons:
This PR improves the `equal_in_freeband` method to work using constant references. It also fixes a bug in the `./etc/test-code-coverage.sh` which prevents it from working on my Ubuntu system.


james-d-mitchell:
Something is wonky in the `doc` job, hopefully that will resolve itself soon.

PR #371, opened 2022-09-15
Add `FullTransformationMonoid` and `PartialTransformationMonoid` presentations

MTWhyte:
This PR adds a constructions for presentations of the full and partial transformation monoids.

### DIFF
--- a/tests/fpsemi-examples.cpp
+++ b/tests/fpsemi-examples.cpp
@@ -1348,15 +1348,14 @@ namespace libsemigroups {
   // See: The Chinese Monoid - Cassaigne, Espie, Krob, Novelli and Hivert, 2001
   std::vector<relation_type> ChineseMonoid(size_t n) {
     std::vector<relation_type> result;
-      for (size_t a = 0; a < n; a++) {
-        for (size_t b = a; b < n; b++) {
-          for (size_t c = b; c < n; c++) {
-            if (a != b) {
-              result.emplace_back(word_type({c, b, a}), word_type({c, a, b}));
-            }
-            if (b != c) {
-              result.emplace_back(word_type({c, b, a}), word_type({b, c, a}));
-            }
+    for (size_t a = 0; a < n; a++) {
+      for (size_t b = a; b < n; b++) {
+        for (size_t c = b; c < n; c++) {
+          if (a != b) {
+            result.emplace_back(word_type({c, b, a}), word_type({c, a, b}));
+          }
+          if (b != c) {
+            result.emplace_back(word_type({c, b, a}), word_type({b, c, a}));
           }
         }
       }
--- a/tests/fpsemi-examples.cpp
+++ b/tests/fpsemi-examples.cpp
@@ -1296,7 +1296,7 @@ namespace libsemigroups {
               {{3, 1, 1, 2, 0}, {3, 1, 1, 2}},
               {{3, 1, 1, 2, 1}, {3, 1, 0, 2}},
               {{1, 2, 1, 2, 0, 2}, {2, 1, 2, 0, 2}}};
-    } else if (n >= 4 && val == author::Aizenstat) {
+    } else if (n >= 4 && val == author::Sutov) {
       // From Theorem 9.4.1, p169, (Ganyushkin + Mazorchuk)
       // https://link.springer.com/book/10.1007/978-1-84800-281-4
       auto result = SymmetricInverseMonoid(n, author::Sutov);
--- a/tests/fpsemi-examples.cpp
+++ b/tests/fpsemi-examples.cpp
@@ -99,7 +99,7 @@ namespace libsemigroups {
         }
       }
     }
-    void add_Iwahori_full_transformation_monoid_relations(
+    void add_iwahori_full_transformation_monoid_relations(
         std::vector<relation_type>& result,
         size_t                      n,
         size_t                      m) {
@@ -112,7 +112,7 @@ namespace libsemigroups {
       // The argument m specifies the letter value the idempotent e12
       // corresponds to. When adding these relations for the full transformation
       // monoid presentation, we want m = n - 1. For the partial transformation
-      // monoid presentaiton, we want m = n.
+      // monoid presentation, we want m = n.
 
       // It is useful to have this as a separate function, to avoid computing
       // duplicate presentations for the underlying symmetric group, when
@@ -129,9 +129,9 @@ namespace libsemigroups {
       // generating set
       auto eps = [&e12, &pi](size_t i, size_t j) -> word_type {
         LIBSEMIGROUPS_ASSERT(i != j);
-        if (i == 1 and j == 2) {
+        if (i == 1 && j == 2) {
           return e12;
-        } else if (i == 2 and j == 1) {
+        } else if (i == 2 && j == 1) {
           return pi[0] * e12 * pi[0];
         } else if (i == 1) {
           return pi[0] * pi[j - 2] * pi[0] * e12 * pi[0] * pi[j - 2] * pi[0];
@@ -149,9 +149,7 @@ namespace libsemigroups {
       auto transp = [&pi](size_t i, size_t j) -> word_type {
         LIBSEMIGROUPS_ASSERT(i != j);
         if (i > j) {
-          size_t k = i;
-          i        = j;
-          j        = k;
+          std::swap(i, j);
         }
         if (i == 1) {
           return pi[j - 2];
@@ -175,7 +173,7 @@ namespace libsemigroups {
           result.emplace_back(transp(i, j) * eps(i, j) * transp(i, j),
                               eps(j, i));
           for (size_t k = 1; k <= n; ++k) {
-            if (k == i or k == j) {
+            if (k == i || k == j) {
               continue;
             }
             // Relations (h)
@@ -193,7 +191,7 @@ namespace libsemigroups {
             result.emplace_back(transp(j, k) * eps(i, j) * transp(j, k),
                                 eps(i, k));
             for (size_t l = 1; l <= n; ++l) {
-              if (l == i or l == j or l == k) {
+              if (l == i || l == j || l == k) {
                 continue;
               }
               // Relations (e)
@@ -1235,7 +1233,7 @@ namespace libsemigroups {
       // of the pi_i and e_12.
       // https://link.springer.com/book/10.1007/978-1-84800-281-4
       auto result = SymmetricGroup(n, author::Carmichael);
-      add_Iwahori_full_transformation_monoid_relations(result, n, n - 1);
+      add_iwahori_full_transformation_monoid_relations(result, n, n - 1);
       return result;
     }
     LIBSEMIGROUPS_EXCEPTION("not yet implemented");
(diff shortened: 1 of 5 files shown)


MTWhyte:
If I run `REQUIRE(p.rules != p.rules)` before creating the ToddCoxeter on line `2628` of `tests/test-todd-coxeter.cpp`, I retrieve the following (modulo formatting) for `p.rules` in the error output:

```
{ { 0, 0 }, { 5 },
  { 1, 1 }, { 5 },
  { 2, 2 }, { 5 },
  { 0, 1, 0, 1, 0, 1 }, { 5 },
  { 1, 2, 1, 2, 1, 2 }, { 5 },
  { 2, 0, 2, 0, 2, 0 }, { 5 },
  { 0, 1, 0, 2, 0, 1, 0, 2 }, { 5 },
  { 1, 2, 1, 0, 1, 2, 1, 0 }, { 5 },
  { 2, 0, 2, 1, 2, 0, 2, 1 }, { 5 },
  { 3, 3 }, { 3 },
  { 3, 0, 3, 0 }, { 0, 3, 0, 3 },
  { 0, 3, 0, 1 }, { 1, 0, 3, 0 },
  { 1, 3, 1, 0 }, { 0, 1, 3, 1 },
  { 0, 3, 0, 2 }, { 2, 0, 3, 0 },
  { 2, 3, 2, 0 }, { 0, 2, 3, 2 },
  { 0, 3, 0, 3, 0 }, { 0, 3, 0, 3 },
  { 4 }, { 4, 4 },
  { 4, 0, 3, 0 }, { 4 },
  { 0, 3, 0, 4 }, { 0, 3, 0 },
  { 4, 3 }, { 0, 3, 0, 3, 4 },
  { 4, 1, 3, 1 }, { 1, 3, 1, 4 },
  { 0, 5 }, { 0 },
  { 5, 0 }, { 0 },
  { 1, 5 }, { 1 },
  { 5, 1 }, { 1 },
  { 2, 5 }, { 2 },
  { 5, 2 }, { 2 },
  { 3, 5 }, { 3 },
  { 5, 3 }, { 3 },
  { 4, 5 }, { 4 },
  { 5, 4 }, { 4 },
  { 5, 5 }, { 5 } }
```

As relates to Theorem 9.4 of *Classical Finite Transformation Semigroups*, the generators `0`, `1`, `2`, `3`, `4` and `5` correspond to pi_2, pi_3, pi_4, epsilon_(1), epsilon_{1, 2} and the identity, respectively. I have thoroughly checked the output, and I'm very confident that it is what it should be.

The relation `{{4, 4}, {4}}` (that the non-permutation full-transformation generator is idempotent) is omitted from the theorem, but I think it's clear that it should be there (and even if I was mistaken, adding that relation couldn't cause the problem).

MTWhyte:
To be a bit more specific: the number of active cosets after about a minute is hovering at around 6 million, which is why I think there is a problem.

tests/fpsemi-examples.cpp
```diff
@@ -1052,17 +1052,17 @@ namespace libsemigroups {
               {{3, 1, 1, 2, 1}, {3, 1, 0, 2}},
               {{1, 2, 1, 2, 0, 2}, {2, 1, 2, 0, 2}}};
     } else if (n >= 4 && val == author::Aizenstat) {
-      LIBSEMIGROUPS_EXCEPTION("not yet implemented");
-      // FIXME this is missing the relations from T_n
```
james-d-mitchell, line 1056: Looking at this quickly just now, have you maybe made the same mistake as me, forgotten to add the relations from the full transformation monoid mentioned in this FIXME?

MTWhyte: Ah I have misread this, in a way I can't even see how I managed to -- I will fix now, thanks!

MTWhyte:
I've now got KnuthBendix computing the correct size of the full transformation monoid of degree 4 in about 200ms. Degree 5 takes roughly a minute, but also gives the correct size.

MTWhyte:
When building the tests, I am warned:

```
tests/fpsemi-examples.cpp:1158:3: warning: non-void function does not return a value in all control paths [-Wreturn-type]
```

I think this means it's possible for a function (either the whole method, or one of the lambdas) whose return type is not `void` to return nothing, but I've tried to fix this where I thought it could be possible. I'll keep looking, but note it here!

MTWhyte:
ToddCoxeter is much faster. Compared to KnuthBendix's minute, it computes the correct size of the full transformation monoid of degree 5 in 300ms. For degree 6, the time for the correct size is about 16 seconds.

MTWhyte:
Looks like the linter (at least) isn't happy, though I did run `make format` on this, and double checked just now by running it again and seeing that it didn't change any of the relevant files. I'll see if I can figure this out tomorrow.

tests/fpsemi-examples.cpp
```diff
@@ -1052,20 +1052,108 @@ namespace libsemigroups {
...
+      auto                   result = SymmetricInverseMonoid(n, author::Sutov);
+      word_type              e12    = {n};
+      std::vector<word_type> pi;
+      for (size_t i = 0; i <= n - 2; ++i) {
+        pi.push_back({i});
+      }
+
+      // The following expresses the epsilon idempotents in terms of the
+      // generating set
+      auto eps = [=](size_t i, size_t j) -> word_type {
```
MTWhyte, line 1066:
```suggestion
      auto eps = [&e12, &pi, etc...](size_t i, size_t j) -> word_type {
```

MTWhyte: Done

MTWhyte:
I've added a separate construction for the full transformation monoid, and combined the then-two functions for full transformation monoid presentations into one. The author value for the one I've added is `author::Iwahori`.

tests/fpsemi-examples.cpp
```diff
@@ -99,6 +99,115 @@ namespace libsemigroups {
...
+        } else if (i == 2) {
+          return pi[j - 2] * pi[0] * e12 * pi[0] * pi[j - 2];
+        }
+        return pi[i - 2] * pi[0] * pi[j - 2] * pi[0] * e12 * pi[0] * pi[j - 2]
+               * pi[0] * pi[i - 2];
+      };
+
+      auto transp = [&pi](size_t i, size_t j) -> word_type {
+        LIBSEMIGROUPS_ASSERT(i != j);
+        if (i > j) {  // FIXME Want to do: return transp(j, i);, but can't do
```
MTWhyte, line 151: *Todo:* I'd prefer to achieve the same here with `return transp(j, i)`, but this won't compile. I understand this is something to do with the lambda function not being within its own scope, and I'll figure out what the best way to sort this is.

MTWhyte: I understand that I can fix this by adding `transp` as a third argument to itself, per [this](https://artificial-mind.net/blog/2020/09/12/recursive-lambdas) solution. However, I think this is way less readable than the temporary 'manual' swap I put in, so I won't. I could also get away with not setting `i < j`, but I think doing this makes the relations easier to read.

I'll leave it as is for now.

tests/fpsemi-examples.cpp
```diff
@@ -1106,7 +1226,8 @@ namespace libsemigroups {
   }
 
   // Chinese monoid
-  // See: The Chinese Monoid - Cassaigne, Espie, Krob, Novelli and Hivert, 2001
+  // See: The Chinese Monoid - Cassaigne, Espie, Krob, Novelli and Hivert,
+  // 2001
```
MTWhyte, line 1230: This is an accidental change from where the formatter alters a file I've modified, but in a place where I've not modified it. I can remove this if that's best.

tests/fpsemi-examples.cpp
```diff
@@ -99,6 +99,114 @@ namespace libsemigroups {
...
+        return pi[i - 2] * pi[0] * pi[j - 2] * pi[0] * e12 * pi[0] * pi[j - 2]
+               * pi[0] * pi[i - 2];
+      };
+
+      auto transp = [&pi](size_t i, size_t j) -> word_type {
+        LIBSEMIGROUPS_ASSERT(i != j);
+        if (i > j) {
+          size_t k = i;
+          i        = j;
+          j        = k;
```
james-d-mitchell, line 154:
```suggestion
   std::swap(i, j);
```

james-d-mitchell: With the correct amount of indentation!

tests/fpsemi-examples.cpp
```diff
@@ -99,6 +99,114 @@ namespace libsemigroups {
...
+      std::vector<word_type> pi;
+      for (size_t i = 0; i <= n - 2; ++i) {
+        pi.push_back({i});
+      }
+
+      // The following expresses the epsilon idempotents in terms of the
+      // generating set
+      auto eps = [&e12, &pi](size_t i, size_t j) -> word_type {
+        LIBSEMIGROUPS_ASSERT(i != j);
+        if (i == 1 and j == 2) {
```
james-d-mitchell, line 132:
```suggestion
        if (i == 1 && j == 2) {
```

james-d-mitchell: Although the keywords `and`, `or`, and `not` exist in C++, we use `&&`, `||` and `!` instead.

tests/fpsemi-examples.cpp
```diff
@@ -99,6 +99,114 @@ namespace libsemigroups {
...
+      // This function adds the full transformation monoid relations due to
+      // Iwahori, from Section 9.3, p161-162, (Ganyushkin + Mazorchuk),
+      // expressed in terms of the generating set {pi_2, ..., pi_n,
+      // epsilon_{12}} using the notation of that chapter.
+      // https://link.springer.com/book/10.1007/978-1-84800-281-4
+
+      // The argument m specifies the letter value the idempotent e12
+      // corresponds to. When adding these relations for the full transformation
+      // monoid presentation, we want m = n - 1. For the partial transformation
+      // monoid presentaiton, we want m = n.
```
james-d-mitchell, line 115:
```suggestion
      // monoid presentation, we want m = n.
```

tests/fpsemi-examples.cpp
```diff
@@ -99,6 +99,114 @@ namespace libsemigroups {
         }
       }
     }
+    void add_Iwahori_full_transformation_monoid_relations(
```
james-d-mitchell, line 102:
```suggestion
    void add_iwahori_full_transformation_monoid_relations(
```
We use snake case for functions (mostly) which doesn't include caps.

tests/fpsemi-examples.cpp
james-d-mitchell: Yes please.

james-d-mitchell (approved):
Couple of minor things to fix, then I'm happy to merge this!

Thanks! @MTWhyte

tests/fpsemi-examples.cpp
MTWhyte: Done

tests/fpsemi-examples.cpp
MTWhyte: Done

tests/fpsemi-examples.cpp
MTWhyte: Done

tests/fpsemi-examples.cpp
MTWhyte: Done

tests/fpsemi-examples.cpp
MTWhyte: Done

PR #379, opened 2022-09-27
Fix typo in a `ToddCoxeter` test

MTWhyte:
This PR fixes a typo in a `ToddCoxeter` extreme test. The number of generators should be three, rather than four, otherwise the congruence has infinitely many classes.

(The test took 5m 55s to run for me, so I also changed the comment with the expected runtime. I know this is hardware dependent, though, so this might not be a good change.)


james-d-mitchell:
Thanks @MTWhyte, good catch!

Issue #383, opened 2022-09-30
Issue with `replace_subword`


MTWhyte:
The presentation helper `replace_subword` doesn't behave properly with the empty word.

Where, say, `presentation::replace_subword(pres, {}, {n - 1}` is called, the the program compiles, but runs for a while and then gets killed.

(I can't assign myself, but I am going to fix this)

james-d-mitchell:
Resolved in v2.3.1

PR #384, opened 2022-09-30
Fix issue with `replace_subword` and the empty word

MTWhyte:
This PR causes an exception to be thrown if the presentation helper `replace_subword` is used on the empty word, addressing issue #383.

### DIFF
--- a/include/libsemigroups/present.tpp
+++ b/include/libsemigroups/present.tpp
@@ -454,7 +454,7 @@ namespace libsemigroups {
                          W const&         existing,
                          W const&         replacement) {
       if (existing.empty()) {
-        LIBSEMIGROUPS_EXCEPTION("the second argument must not be the empty word");
+        LIBSEMIGROUPS_EXCEPTION("the second argument must not be empty");
       }
       auto rplc_sbwrd = [&existing, &replacement](W& word) {
         auto it = std::search(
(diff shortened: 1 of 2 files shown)


include/libsemigroups/present.tpp
```diff
@@ -453,6 +453,9 @@ namespace libsemigroups {
     void replace_subword(Presentation<W>& p,
                          W const&         existing,
                          W const&         replacement) {
+      if (existing.cbegin() == existing.cend()) {
```
MTWhyte, line 456: I originally tried `if (existing == word_type({}))`, but there was a type mismatch I wasn't confident about fixing so I did it this way. I'm completely open to suggestion here, since I don't think this is as elegant as could be!

james-d-mitchell: Should probably be `if (existing.empty()) {`

james-d-mitchell:
Be good to have some tests for this too @MTWhyte

james-d-mitchell:
I wrote too soon, you've already added a test, amazing !

include/libsemigroups/present.tpp
```diff
@@ -453,6 +453,9 @@ namespace libsemigroups {
     void replace_subword(Presentation<W>& p,
                          W const&         existing,
                          W const&         replacement) {
+      if (existing.empty()) {
+        LIBSEMIGROUPS_EXCEPTION("the second argument must not be the empty word");
```
james-d-mitchell, line 457:
```suggestion
        LIBSEMIGROUPS_EXCEPTION("the second argument must not be empty");
```

Issue #391, opened 2022-10-06
Out of bounds vector access


jamesjer:
I attempted to build libsemigroups 2.3.0 for Fedora.  The Fedora build flags include `-D_GLIBCXX_ASSERTIONS`, which I highly recommend for testing builds.  One test failed:
```
-------------------------------------------------------------------------------
Sims1 000: fp example 1
-------------------------------------------------------------------------------
tests/test-sims1.cpp:77
...............................................................................

tests/test-sims1.cpp:77: FAILED:
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal

Sims1 000: fp example 1 . . .                                                  
===============================================================================
test cases:     869 |     868 passed | 1 failed
assertions: 1162775 | 1162774 passed | 1 failed

/usr/include/c++/12/bits/stl_vector.h:1123: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = libsemigroups::Sims1<unsigned int>::PendingDef; _Alloc = std::allocator<libsemigroups::Sims1<unsigned int>::PendingDef>; reference = libsemigroups::Sims1<unsigned int>::PendingDef&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
make: *** [Makefile:5233: check] Aborted (core dumped)
```

Weirdly, I cannot get this to happen with GDB, but it happens every time when running the test from the command line.  Valgrind, on the other hand, provided a backtrace:
```
==68== Process terminating with default action of signal 6 (SIGABRT)
==68==    at 0x4D2D2EC: __pthread_kill_implementation (pthread_kill.c:44)
==68==    by 0x4CDB005: raise (raise.c:26)
==68==    by 0x4CC47FB: abort (abort.c:79)
==68==    by 0x4A46EAF: std::__glibcxx_assert_fail(char const*, int, char const*, char const*) (debug.cc:60)
==68==    by 0x95A7EE: UnknownInlinedFun (stl_vector.h:1123)
==68==    by 0x95A7EE: UnknownInlinedFun (stl_vector.h:1121)
==68==    by 0x95A7EE: UnknownInlinedFun (sims1.tpp:707)
==68==    by 0x95A7EE: UnknownInlinedFun (sims1.tpp:719)
==68==    by 0x95A7EE: UnknownInlinedFun (sims1.tpp:790)
[...]

PR #392, opened 2022-10-07
Fix another Sims1 corner case, resolve issue 391, and add some tests

james-d-mitchell:


james-d-mitchell:
I think this resolves #391, @jamesjer any chance you could confirm?

jamesjer:
This does indeed fix the issue.  Thank you!

PR #394, opened 2022-10-10
Fix inaccurate exception information for replace_subword

MTWhyte:
This PR fixes an inaccuracy in the documentation of the presentation helper `replace_subword`, introduced by the changes in #384.

### DIFF
--- a/include/libsemigroups/present.hpp
+++ b/include/libsemigroups/present.hpp
@@ -761,8 +761,7 @@ namespace libsemigroups {
     //!
     //! \returns (None)
     //!
-    //! \exceptions
-    //! \no_libsemigroups_except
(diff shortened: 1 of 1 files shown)


include/libsemigroups/present.hpp
```diff
@@ -761,8 +761,7 @@ namespace libsemigroups {
     //!
     //! \returns (None)
     //!
-    //! \exceptions
-    //! \no_libsemigroups_except
+    //! \throws LibsemigroupsException if `existing` is the empty word.
```
james-d-mitchell, line 764:
```suggestion
    //! \throws LibsemigroupsException if `existing` is empty.
```

PR #414, opened 2022-11-19
build: add missing eigen3 to pkg-config file

jengelh:
A libsemigroups installed to /usr requires eigen3 to be present due to an #include line,

include/libsemigroups/digraph.hpp:#include <Eigen/Core>

However, with eigen3 not being specified in the .pc file, package managers do not install eigen3 alongside libsemigroups-devel. Rectify that.


james-d-mitchell:
Thanks for this @jengelh, much appreciated. The CI for the `libsemigroups_pybind11` seems to fail, not sure if that's an issue here or there? Any thoughts?

james-d-mitchell:
Awesome thanks so much @jengelh

PR #419, opened 2022-11-22
Put `fpsemi-examples` in the namespace `fpsemigroup`

MTWhyte:
Per earlier conversation with @james-d-mitchell, this PR moves the `fpsemi-examples` functions into the `fpsemigroup` namespace.

Issue #411, opened 2022-11-17
Add non-examples of presentations


james-d-mitchell:
Candidates include:

- The Godelle presentations for Renner monoids
- The presentation for the symmetric group with author value `author::Guralnick + author::Kantor + author::Kassabov + author::Lubotzky`

PR #421, opened 2022-11-23
Add a `not_symmetric_group` presentation

MTWhyte:
**Note: This should be merged after #419**

This PR adds a function `not_symmetric_group` presentation, which returns a presentation claiming to define the symmetric group (but it does not).

This relates to #411.


james-d-mitchell:
Thanks @MTWhyte

PR #431, opened 2022-12-12
Add a second symmetric group presentation due to Moore

MTWhyte:
This PR adds a second symmetric group presentation due to Moore, and therefore sets up the `symmetric_group` function to receive a third argument called `index` which can be used to choose between presentations of the same author.

In a separate commit, default value tests for the function `cyclic_inverse_monoid` are added.

### DIFF
--- a/include/libsemigroups/fpsemi-examples.hpp
+++ b/include/libsemigroups/fpsemi-examples.hpp
@@ -285,36 +285,35 @@ namespace libsemigroups {
     //! Returns a vector of relations giving a monoid presentation for the
     //! symmetric group. The argument `val` determines the specific presentation
     //! which is returned. The options are:
-    //! * `author::Burnside + author::Miller` (given on p.464 of
-    //! [10.1017/CBO9781139237253][])
-    //! * `author::Carmichael` (given in comment 9.5.2 of
-    //! [10.1007/978-1-84800-281-4])
-    //! * `author::Coxeter + author::Moser` (see Ch. 3, Prop 1.2 of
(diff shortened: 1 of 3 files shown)


MTWhyte:
@james-d-mitchell a particular question I have is: do you think the way I've documented `index` is clear enough? I've tried to get across that an exception will be thrown if `index` is out of bounds, without writing a new`\throws LibsemigroupsException` specific to every single choice of author.

MTWhyte:
I also considered adding the numbers of generators and relations for the symmetric group presentations while I was doing this. Since there are quite a lot of presentations (more than for `cyclic_inverse_monoid` and `order_preserving_monoid`), I wonder whether instead of having the list of bullet points with citations in brackets, we should have a table with fields _author_, _index_, _reference_, _number of generators_ and _number of relations_?

james-d-mitchell:
> I wonder whether instead of having the list of bullet points with citations in brackets, we should have a table with fields _author_, _index_, _reference_, _number of generators_ and _number of relations_?

That'd be great, good idea.

james-d-mitchell:
Also @MTWhyte would you mind rebasing this onto `main`, I made some changes to the CI. Thanks

MTWhyte:
I've got the table working properly now. Including `NOLINT` as an html comment seems to have worked. A problem worth noting for future reference is, it appears that if the code for the table contains a line longer than about 160 characters, the table doesn't build properly.

james-d-mitchell:
Thanks @MTWhyte

PR #433, opened 2022-12-15
Add presentation for the monoid of partial isometries of a cycle graph

MTWhyte:
This PR adds a presentation for the monoid of partial isometries of a cycle graph, as described in Theorem 2.8 [here](https://arxiv.org/pdf/2205.02196.pdf).

### DIFF
--- a/docs/source/api/fpsemi-examples.rst
+++ b/docs/source/api/fpsemi-examples.rst
@@ -23,7 +23,7 @@ Presentations from the following sources are implemented: :cite:`Gay1999aa`;
 :cite:`East2011aa`; :cite:`Ayik2000aa`; :cite:`Ruskuc1995aa`;
 :cite:`Aizenstat1958aa`; :cite:`Coxeter1979aa`; :cite:`Knuth1970aa`;
 :cite:`Lascoux1981aa`; :cite:`Moore1897aa`; :cite:`Aizenstat1962aa`;
-:cite:`Fernandes2022aa`.
+:cite:`Fernandes2022aa`; :cite:`Fernandes2022ab`.
 
 .. cpp:type:: libsemigroups::fpsemigroup::author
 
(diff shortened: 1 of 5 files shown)


james-d-mitchell:
@MTWhyte  I think you'll need to mark some of the tests as `no-valgrind` and `no-coverage` and to add some tests for the exceptions.

MTWhyte:
Will do, for valgrind!

For the coverage, I've got a test for the exceptions on line `383` already (test case `069`). If I look in 'Files changed' on here, it looks like the code coverage is complaining about tens of code blocks that haven't been modified by this PR.

docs/source/api/fpsemi-examples.rst
```diff
@@ -23,7 +23,7 @@ Presentations from the following sources are implemented: :cite:`Gay1999aa`;
 :cite:`East2011aa`; :cite:`Ayik2000aa`; :cite:`Ruskuc1995aa`;
 :cite:`Aizenstat1958aa`; :cite:`Coxeter1979aa`; :cite:`Knuth1970aa`;
 :cite:`Lascoux1981aa`; :cite:`Moore1897aa`; :cite:`Aizenstat1962aa`;
-:cite:`Fernandes2022aa`.
+:cite:`Fernandes2022aa`; :cite:`Fernandes2022ab`.
```
james-d-mitchell, line 26:
```suggestion
:cite:`Fernandes2022aa`; :cite:`Fernandes2022ab`. 
```

james-d-mitchell:
Thanks for the PR @MTWhyte I think you need to add the `no-valgrind` tag to a bunch of tests though, since there are several that run for 30+ seconds. Thanks!

james-d-mitchell (changes requested):
I'm happy with this except for the failing val grind ci job that can be fixed by adding the `no-valgrind` tag to the tests that run for more than a couple of seconds.

MTWhyte:
Thanks @james-d-mitchell, I've added the `no-valgrind` tags now to the appropriate tests (it's perhaps worth noting that I didn't add or modify these tests as part of this PR, but I imagine that's normal enough).

james-d-mitchell:
Thanks @MTWhyte feel free to ignore the coverage warnings I'm not sure what they're all about.

Issue #438, opened 2022-12-20
Add `presentation::make_semigroup_presentation`


james-d-mitchell:
Basically just wrapping:

```
presentation::replace_word(p, {}, {p.alphabet().size()});
presentation::add_identity_rules(p, p.alphabet().size());
p.alphabet(p.alphabet().size() + 1);
```

or equivalent.

MTWhyte:
I will do this

james-d-mitchell:
Resolved in v2.6.0

PR #442, opened 2023-01-20
Change `orientation_preserving_monoid` and `orientation_reversing_monoid` to give monoid presentations

MTWhyte:
This PR modifies the output of the functions `orientation_preserving_monoid` and `orientation_reversing_monoid` to give monoid presentations rather than semigroup presentations, as part of the standardisation process discussed with @james-d-mitchell.

(I was tempted to address #438 simultaneously, but it was tricky to get it working for words and strings in the way I wanted so I'll do this separately).


MTWhyte:
(I forgot these functions existed beforehand, so missed a lot of places where they come up in the tests! I'll fix this)

MTWhyte:
Note: In this PR I will change the CI file to use a version of `libsemigroups_pybind11` where I've changed it to work properly with these `libsemigroups` changes.

james-d-mitchell:
This didn't run the azure jobs, because they're only enabled for pull requests into `stable-*.*` and `main`:

https://github.com/libsemigroups/libsemigroups/blob/da5a9f430cc4400aea186ae0cbaeb18d2ce55f83/azure-pipelines.yml#L1-L3

if you add `- v3` to the list of branches then the azure CI jobs will trigger too.

james-d-mitchell:
The CI seems to be suffering some issues at present, I'll re-run the failing test (when it fails which it looks like it will) in the morning, hopefully the problem will have cleared by then.

PR #444, opened 2023-01-31
froidure-pin: use `ActionDigraph` rather than `DynamicArray2`

james-d-mitchell:


james-d-mitchell:
Closing this in because the changes are now included in the forthcoming v3

PR #450, opened 2023-02-04
word: remove `detail::StringToWord`

james-d-mitchell:
This is only partially done, I'd like to actually remove/rename `StringToWord` and `word_to_string` and replace them with `make`


james-d-mitchell:
Requires some doc

james-d-mitchell:
I think I decided to retain some version of `StringToWord`

PR #455, opened 2023-02-06
Add `congruence_kind` argument check to `ToddCoxeter` function `to_gap_string`

MTWhyte:
This PR adds an argument check to the `ToddCoxeter` function `to_gap_string`, to require that this only be done where the `ToddCoxeter` instance is `twosided`.


james-d-mitchell (approved):
Great, thanks @MTWhyte, I'll merge this as soon as the CI completes

PR #457, opened 2023-02-07
Add functions useful for manipulating words

MTWhyte:
This PR adds some functions which are useful for manipulating words. Currently it is a draft, as there are a few more things to do, including:

- [x] Make a `prod` method which uses `insert`, where step is `1` or `-1`.
- [x] Make a method which takes a word by reference and puts the output of `prod` into that word
- [x] In `prod`, get `abs` working so I can run `result.reserve(<appropriate size>)`
- [x] Update the tests to use the new `_w` feature added in #449
- [x] Move things to a `.cpp` file?
- [x] Get the documentation to work


james-d-mitchell:
Maybe remove the white space only changes altogether, or put them in a separate commit too? (I.e. looks like some of the comment lines in `present.hpp` that are unrelated to this PR have been re-wrapped). Please make sure you're using clang-format v15. 

Also: I think we should put these functions into the namespace `presentation`, and we'll have to add some template voodoo to avoid installing `operator<<` for any type whatsoever (I can give you some help with this last bit).

MTWhyte:
Ah sorry that must've crept in when I rebased, I've ditched the wrapping changes now.

docs/source/api/word-functions.rst
```diff
@@ -0,0 +1,41 @@
...
+Contents
+~~~~~~~~
+
+.. cpp:namespace:: libsemigroups::fpsemigroup
+
+.. list-table::
+   :widths: 50 50
+   :header-rows: 0
+
+   * - :cpp:any:`operator<<`
```
james-d-mitchell, line 26: I think you'll need to add the arguments to this to make the doc work, there should be plenty of similar examples in the doc already.

MTWhyte:
I think this is now in reviewable shape, other than the problem with `doxygen`'s `\ref` for overloaded functions. Cygwin also doesn't seem to like `insert_prod` at compilation, but I can't figure out from the logs why this happens.

james-d-mitchell:
@MTWhyte try replacing:

~~~
90 |       w.insert(w.cend(), u.cbegin(), u.cend());
~~~

with 

~~~
90 |       w.insert(w.end(), u.cbegin(), u.cend());
~~~

The "c" indicates `const`, which might mean you can't modify it, that should fix the cygwin ci.

docs/source/api/present-helper.rst
```diff
@@ -30,6 +30,18 @@ Contents
...
+   * - :cpp:any:`operator<<`
+     - Concatenate two words or strings.
+
+   * - :cpp:any:`pow`
+     - Take a power of a word or string.
+
+   * - :cpp:any:`prod`
+     - Take a product from a collection of letters.
+
+   * - :cpp:any:`insert_prod`
```
james-d-mitchell, line 42: I'm not convinced that `insert_prod` is actually useful, more later.

include/libsemigroups/present.hpp
```diff
@@ -577,7 +697,7 @@ namespace libsemigroups {
 
     //! Add rules for an identity element.
     //!
-    //! Adds rules of the form \f$ae = ea = a\f$ for every letter \f$a\f$ in
```
james-d-mitchell, line 580: Good catch!

include/libsemigroups/present.tpp
```diff
@@ -20,6 +20,76 @@
 // monoid presentations.
 
 namespace libsemigroups {
+  namespace presentation {
+
+    template <typename T>
+    T operator<<(T const& u, T const& w) {
```
james-d-mitchell, line 26: This needs to check whether or not `T` satisfies `IsWord` as in `operator<<=` below

include/libsemigroups/present.hpp
```diff
@@ -46,6 +46,126 @@
...
+    template <typename T>
+    void insert_prod(T&                            w,
+                     std::initializer_list<size_t> ilist,
+                     size_t                        first,
+                     size_t                        last,
+                     int                           step) {
+      insert_prod(w, word_type(ilist), first, last, step);
+    }
+  }  // namespace presentation
+
```
james-d-mitchell, line 168: The declaration and doc for `<<=` is missing

include/libsemigroups/present.tpp
```diff
@@ -20,6 +20,76 @@
...
+      return result;
+    }
+
+    template <typename T, typename = std::enable_if_t<IsWord<T>::value>>
+    void operator<<=(T& u, T const& v) {
+      u.insert(u.end(), v.cbegin(), v.cend());
+    }
+
+    template <typename T>
+    T pow(T const& x, size_t n) {
```
james-d-mitchell, line 38: I'd add a static assertion here that `T` satisfies `IsWord`

include/libsemigroups/present.tpp
```diff
@@ -20,6 +20,76 @@
...
+        if (n % 2 == 1) {
+          z <<= y;
+        }
+      }
+      return z;
+    }
+
+    // Note: we could do a version of the below using insert on words, where
+    // the step is +/- 1.
+    template <typename T>
```
james-d-mitchell, line 55: I'd add a static assertion here too

include/libsemigroups/present.tpp
```diff
@@ -20,6 +20,76 @@
...
+      result.insert(result.end(), w.cbegin(), w.cend());
+      return result;
+    }
+
+    template <typename T, typename = std::enable_if_t<IsWord<T>::value>>
+    void operator<<=(T& u, T const& v) {
+      u.insert(u.end(), v.cbegin(), v.cend());
+    }
+
+    template <typename T>
```
james-d-mitchell, line 37: Be good to have a version of this that changed its first argument in-place (i.e. so that the signature is `T&`)

include/libsemigroups/present.tpp
```diff
@@ -20,6 +20,76 @@
...
+        size_t steppos = static_cast<size_t>(-step);
+        for (size_t i = first; i > last; i -= steppos) {
+          result.push_back(elts[i % elts.size()]);
+        }
+      }
+      return result;
+    }
+
+    template <typename T>
+    void insert_prod(T& w, T const& elts, size_t first, size_t last, int step) {
```
james-d-mitchell, line 88: I think this can be removed completely.

src/present.cpp
```diff
@@ -0,0 +1,46 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
+
+// This file contains
```
james-d-mitchell, line 19: This is incomplete.

src/present.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+    std::string pow(char const* w, size_t n) {
+      return pow(std::string(w), n);
+    }
+    word_type pow(std::initializer_list<size_t> ilist, size_t n) {
+      return pow(word_type(ilist), n);
+    }
+
+    std::string operator<<(std::string const& u, char const* w) {
+      return std::string(u) << std::string(w);
```
james-d-mitchell, line 36:
```suggestion
      return u << std::string(w);
```

tests/test-present.cpp
```diff
@@ -1395,4 +1396,109 @@ namespace libsemigroups {
...
+
+    REQUIRE((word_type({0, 1, 0}) << word_type({2}))
+            == word_type({0, 1, 0, 2}));
+    REQUIRE((word_type({0}) << word_type({})) == word_type({0}));
+    REQUIRE((word_type({}) << word_type({0})) == word_type({0}));
+
+    std::string x = "ab";
+    std::string y = "c";
+    REQUIRE((x << y) == "abc");
+    REQUIRE((std::string("") << "a") == "a");
```
james-d-mitchell, line 1418: Be good to have a version of `<<` for `char const* and `char const*` so that this could just be `"" << "a"`. Now that I'm writing this, maybe `<<` should rather be `+` because then we could just do `"" + "a"` and there'd be nothing to implement at all for strings (only for `word_type`), which would also remove the template code, and generally simplify this.

tests/test-present.cpp
```diff
@@ -1395,4 +1396,109 @@ namespace libsemigroups {
...
+                          "[quick][word_functions]") {
+    using namespace literals;
+    using presentation::pow;
+    word_type w = 01_w;
+    REQUIRE(pow(w, 0) == word_type({}));
+    REQUIRE(pow(w, 1) == w);
+    REQUIRE(pow(w, 2) == 0101_w);
+    REQUIRE(pow(pow(w, 2), 3) == 010101010101_w);
+    for (size_t i = 0; i <= 1'000'000; i += 1000) {
+      REQUIRE(pow(word_type({0}), i) == word_type(i, 0));
```
james-d-mitchell, line 1436:
```suggestion
      REQUIRE(pow(0_w, i) == word_type(i, 0));
```

tests/test-present.cpp
```diff
@@ -1395,4 +1396,109 @@ namespace libsemigroups {
...
+    insert_prod(w3, eps, 5, 0, -1);
+    REQUIRE(w3 == 0154321_w);
+
+    word_type w4 = 00_w;
+    insert_prod(w4, {1, 2, 4, 5}, 0, 8, 3);
+    REQUIRE(w4 == 00154_w);
+
+    std::string eps2 = "abcdef";
+    std::string u    = "a";
+    insert_prod(u, eps2, 0, 6, 2);
```
james-d-mitchell, line 1497:
```suggestion
    u <<= prod(eps2, 0, 6, 2);
```

james-d-mitchell (changes requested):
Looks like a good start, plenty of comments to address. Maybe the biggest is: should we maybe rename `<<` to `+` so that we just don't have to implement these for strings, and we can avoid the template stuff? I didn't think of this until I was reading your PR.

include/libsemigroups/present.hpp
```diff
@@ -46,6 +46,126 @@
...
+    //!
+    //! \noexcept
+    template <typename T>
+    T operator<<(T const& u, T const& w);
+
+    std::string operator<<(std::string const& u, char const* w);
+
+    //! Take a power of a word or string.
+    //!
+    //! Returns the `n`th power of the word/string given by `u` .
```
MTWhyte, line 82:
```suggestion
    //! Returns the `n`th power of the word/string given by `w` .
```

MTWhyte:
Todo: Update documentation for changes I've made to signatures.

MTWhyte:
I believe all that remains is getting the hyperlinks to work. I have tried (without success) many things including:
- Various permutations of namespaces and function signatures after `\ref`
- Using the `\fn ...` command at the beginning of the documentation; and various namespace and signature permutations of this
- Attempting to recover `doxygen`'s internal name for this, and 'demangling' it
- Adding a hyperlink in plain `HTML` (this showed up as a link, but it pointed to nowhere)

MTWhyte:
To note: the table of contents links for `pow`, `prod` and `prod_inplace` seem to no longer be working. In the doc source file they're written in the way I thought the solution was, so I don't think this is the consequence of merging error.

MTWhyte:
This is now part of the Todd-Coxeter refactoring.

PR #471, opened 2023-03-13
Add Catarino presentation to `orientation_preserving_monoid`

MTWhyte:
This PR adds a presentation due to Catarino to the `orientation_preserving_monoid` function, as mentioned [in this paper](https://link.springer.com/article/10.1007/s10012-000-0001-1).

This remains as a draft, since there's documentation to sort out. I've held off doing this for now since there'll be merge conflicts -- this PR is currently based on #457, and it really should have a relationship to #442 (since this changes the existing to give monoid rather than semigroup presentations, and that should be consistent between different authors for the same function).


src/fpsemi-examples.cpp
```diff
@@ -937,28 +944,72 @@ namespace libsemigroups {
...
+        }
+
+        // rules 10
+        for (size_t i = 0; i <= n - 3; ++i) {
+          result.emplace_back(v[i + 1] + a, a + v[i]);
+        }
+
+        // rule 11
+        result.emplace_back(u[n - 2] + a,
+                            pow(a, 2) + prod(v, n - 2, 0, -1) + v[0]);
```
MTWhyte, line 998: @james-d-mitchell this PR also suggests that `prod` might want to take an `int` argument for its upper and lower limits, rather than `size_t`. Since we can't have `-1` as the third argument to `prod`, it can't include the `0`th entry of its first argument if we want a product in reverse order.

james-d-mitchell: Good point! I agree, it should be an `int`

MTWhyte:
Update to return a monoid presentation, and make this PR into the refactor Todd coveter branch of James's fork.

MTWhyte:
And also change it to use the new behaviour of `prod`.

james-d-mitchell:
@MTWhyte any chance of resurrecting this PR and making into the v3 branch?

MTWhyte:
I will absorb this into #517 , once I've figured out why the tests on that are currently failing.

PR #473, opened 2023-03-16
More new features for presentations

james-d-mitchell:


reiniscirpons (changes requested):
One small quibble about some implicit assumptions on the number of generators, otherwise looks great and compiles fine on my machine.

PR #476, opened 2023-03-21
Ukkonen: expose suffix tree impl

james-d-mitchell:


reiniscirpons (approved):
Looks good to me, compiles on my system and almost all tests pass.

The only issue is with tests `071` and `072` failing in the `test_kambites` tests. Something that should be zero isn't:
```
Kambites 071: (fpsemi) almost all 2-generated 1-relation monoids are C(4)
tests/test-kambites.cpp:1670

tests/test-kambites.cpp:1672: FAILED:
  REQUIRE( x.first == 0 )
with expansion:
  235629 (0x3986d) == 0
```
However, this test also fails in the main branch of libsemigroups so is likely not caused by the changes of this commit. Since the commit does touch the `kambites.hpp` file, it might make sense to mark these tests as failing here?

PR #483, opened 2023-09-20
Update Konieczny test for faster computers

flsmith:
One of the Konieczny tests checks that the algorithm can be paused and resumed correctly. On newer computers the initial time given to run may be too long, which causes a failure in the test.

While this patch fixes the immediate issue, it's not particularly satisfactory. Any thoughts @james-d-mitchell ?


james-d-mitchell:
@flsmith thanks for this! If you wanted to make the test more robust you could use `run_until` with a predicate, instead of `run_for` which will always be rather fragile. Using `run_until` will test the same behaviour of stopping and resuming the computation later.

james-d-mitchell:
Rebase this after #484 passes and is merged

PR #484, opened 2023-10-05
Simplify GAP ci job

james-d-mitchell:

PR #485, opened 2023-10-05
Fix warnings

james-d-mitchell:
This is a duplicate of @reiniscirpons PR, just to see if the CI will run here.


james-d-mitchell:
I think the weirdness has been caused by the ci using v15, me updating my clang-format to use v17, and then the gh action only having v16 available. Hopefully fixed by this.

james-d-mitchell:
And the GAP ci tests are failing because of 

https://github.com/semigroups/Semigroups/issues/960

james-d-mitchell:
Rebase this after #484 passes and is merged.

PR #490, opened 2023-12-09
configure.ac: fix shell equality test

orlitzky:
There was a == that should be an = because in the shell they aren't equal.


james-d-mitchell:
Thanks @orlitzky much appreciated

PR #504, opened 2024-04-09
words: try to fix bug in prod

james-d-mitchell:


james-d-mitchell:
@MTWhyte it seems that the code for product doesn't work properly, you can see this if you run the code in this branch in debug mode. I added a couple of assertions to the code to check that the indices being used are within bounds, and unfortunately they aren't:
~~~
❯ ./test_words "[035]"
Filters: [035]
Assertion failed: (a < elts.size()), function prod, file words.hpp, line 2128.
[035]: Words 035: prod
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
test_words is a Catch v2.13.10 host application.
Run with -? for options

Words 035: prod
tests/test-words.cpp:1094

tests/test-words.cpp:1125: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal

............
===============================================================================
Total time ....................................................................
===============================================================================
test cases:  1 |  0 passed | 1 failed
assertions: 20 | 19 passed | 1 failed

[1]    88011 abort      ./test_words "[035]"
~~~

Can you please try to fix this? Thanks

PR #517, opened 2024-04-09
V3 work on `fpsemi-examples`

MTWhyte:
So far this checks each function in the file for whether it returns a semigroup or monoid presentation, updates the documentation to reflect this, and makes tests for this.

### DIFF
--- a/benchmarks/bench-todd-coxeter.cpp
+++ b/benchmarks/bench-todd-coxeter.cpp
@@ -268,7 +268,7 @@ namespace libsemigroups {
         "OP_n",
         3,
         9,
-        orientation_preserving_monoid,
+        [](size_t n) { return orientation_preserving_monoid(n); },
         "orientation_preserving_monoid",
         strategies,
         DoNothing);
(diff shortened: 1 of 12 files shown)


james-d-mitchell:
@MTWhyte I took the liberty of rebasing this onto the current v3 branch, I hope that's ok!

james-d-mitchell:
I also fixed the linting and formatting.

MTWhyte:
I think this is now is a state where it's worth being reviewed.

I've checked the tests are correctly categorised. On my system, complied with optimisation, no 'quick' test takes longer than 300ms to run (in fact the entire collection of these takes only about 900ms to run).

@james-d-mitchell there are a couple of TODOs, some of which I think will make more sense to you than to me. There are also a couple of presentations I think you've added that don't currently have documentation, but I'm not really sure what any of these actually are so I may leave this for you.

MTWhyte:
The documentation shows up in a sort of weird way with a page I feel is unnecessary. I remember discussing this with @james-d-mitchell a month or so ago, and not being sure how to fix this, so I'll just leave it as is for now.

james-d-mitchell (approved):
Looks great, thanks @MTWhyte

PR #520, opened 2024-04-10
ci: simplify x86-32

james-d-mitchell:


Joseph-Edwards:
This issue seems to suggest this won't work:
https://github.com/actions/runner/issues/2115

james-d-mitchell:
> This issue seems to suggest this won't work:
> [actions/runner#2115](https://github.com/actions/runner/issues/2115)

That is that running the ci job in the actual container (as we are doing in the downstream / GAP / semigroups job) won't work, but somehow hopefully running the container manually might still work?

PR #531, opened 2024-04-17
c3 Further improve Doxygen output

reiniscirpons:
This PR is aimed for the `libsemigroups` v3.0.0 release. It further improves Doxygen output by removing the redundant "Detailed Description" headings, fixing some bugs in the previous navigation bar `fixup` script and adding the ability to add headings in the navigation bar.

@james-d-mitchell This PR should fix the navigation bar not automatically expanding clicked tabs. If not then it may be a Doxygen version issue. I am using `1.10.0`, and cant easily install lower versions. I think you may be on `1.9.8`, could you check out this branch, try compiling the doc and report any bugs you see?


james-d-mitchell:
Thanks @reiniscirpons, I'm not sure the changes in this PR work for me, but I'm also not 100% sure what changes I should expect, let's discuss!

PR #532, opened 2024-04-22
Finalize runner.hpp for v3

james-d-mitchell:


james-d-mitchell:
The pybind11 CI job is expected to fail

PR #537, opened 2024-04-30
v3 Fix bug in `SimsRefinerIdeals` when dealing with complete graphs

reiniscirpons:
This PR fixes a bug, which causes ideal finding to return the wrong answer. In particular, `SimsRefinerIdeals` returned `true` if the input word graph is complete and no sink is found. Due to the way we implement pruning in the current implementation of `Sims`, pruners should always correctly accept or reject a complete word graph (whereas for incomplete word graphs we only need to correctly reject).

To fix this, `SimsRefinerIdeals` will now return `false` when given as input a complete word graph without a nontrivial sink. Note that when given a complete word graph with a nontrivial sink, the remaining checks of the function guarantee that the word graph is accepted if and only if it represents a nontrivial right Rees congruence.

### DIFF
--- a/include/libsemigroups/sims.hpp
+++ b/include/libsemigroups/sims.hpp
@@ -324,7 +324,7 @@ namespace libsemigroups {
     //! * \ref number_of_threads to set the number of threads;
     //! * \ref include to set the pairs to be included;
     //! * \ref exclude to set the pairs to be excluded;
-    //! * \ref add_pruner to add a pruninf function;
+    //! * \ref add_pruner to add a pruner;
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
Thanks for the PR @reiniscirpons, it seems that the coverage CI job is timing out. It seemed to be passing just a few days ago, I wonder if some of the tests you add in this PR should have the `[no-coverage]` tag? I had a quick look and I couldn't spot which test that might be, the issue might be something else. Could you please take a look?

PR #542, opened 2024-05-20
Partially fix undefined behaviour

james-d-mitchell:


Joseph-Edwards:
These changes fix `usan` warnings on test 031 on my machine. Thanks!

PR #550, opened 2024-05-31
Move rewriters to detail

Joseph-Edwards:
This PR adds a small amount of doc, and moves the classes defined in `rewriters.hpp` into the `detail` namespace.


include/libsemigroups/knuth-bendix.hpp
```diff
@@ -115,20 +115,21 @@ namespace libsemigroups {
...
     ////////////////////////////////////////////////////////////////////////
 
-    using external_string_type = std::string;
-    using internal_string_type = std::string;
-    using external_char_type   = char;
-    using internal_char_type   = char;
+    // using external_string_type = std::string;
+    // using internal_string_type = std::string;
+    // using external_char_type   = char;
+    // using internal_char_type   = char;
```
Joseph-Edwards, line 121: These were defined both here and in `rewriters.hpp`. I've commented these out for now, and changed all occurrences to be the `detail` version (apart from in `KBE` because at a quick glance that looked like it was doing something different). Is this a reasonable thing to do?

james-d-mitchell:
I think the test failures are expected right? Ready to merge? @Joseph-Edwards

Joseph-Edwards:
> I think the test failures are expected right? Ready to merge? @Joseph-Edwards 

Yeah that's right, though it did remind me that the bindings need to be slightly modified too. I'm happy to merge

PR #551, opened 2024-06-27
Add Presentation

Joseph-Edwards:
This PR includes the `Presentation` class in a state ready for V3. In particular:
- [X] use of `no_checks` functions
- [X] TODOs and FIXMEs reviewed
- [x] doc
- [x] Python bindings
- [x] `iwyu`

PR #555, opened 2024-07-11
Add to_presentation

Joseph-Edwards:
This PR includes the `to_presentation` functions in a state ready for V3. In particular:
- [X] TODOs and FIXMEs reviewed
- [x] doc
- [x] Python bindings
- [x] `iwyu`

### DIFF
--- a/include/libsemigroups/to-presentation.hpp
+++ b/include/libsemigroups/to-presentation.hpp
@@ -1,6 +1,6 @@
 //
 // libsemigroups - C++ library for semigroups and monoids
-// Copyright (C) 2023 - 2024 James D. Mitchell
+// Copyright (C) 2023-2024 James D. Mitchell
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
--- a/include/libsemigroups/to-presentation.hpp
+++ b/include/libsemigroups/to-presentation.hpp
@@ -1,6 +1,6 @@
 //
 // libsemigroups - C++ library for semigroups and monoids
-// Copyright (C) 2023 James D. Mitchell
+// Copyright (C) 2023 - 2024 James D. Mitchell
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
@@ -16,7 +16,7 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 //
 
-// TODO tpp file
+// TODO(later) tpp file
 
 #ifndef LIBSEMIGROUPS_TO_PRESENTATION_HPP_
 #define LIBSEMIGROUPS_TO_PRESENTATION_HPP_
@@ -31,7 +31,14 @@
 
 namespace libsemigroups {
 
-  //! Make presentation from a FroidurePin object.
+  //! \defgroup to_presentation_group Making presentations from other objects
+  //!
+  //! This file contains documentation for creating semigroup and monoid
+  //! presentations in `libsemigroups` from other types of objects.
+
+  //! \ingroup to_presentation_group
+  //!
+  //! \brief Make a presentation from a FroidurePin object.
   //!
   //! This function constructs and returns a Presentation object using the
   //! rules of a FroidurePin object, accessed via FroidurePin::cbegin_rules and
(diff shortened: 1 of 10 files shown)


Joseph-Edwards:
This PR was written assuming the changes in #551 get merged into the v3 branch. If they do, this branch should be rebased on top of the v3 branch, and the commit https://github.com/libsemigroups/libsemigroups/pull/555/commits/5a2644e9f0e1e6c758e09f6f22d76eca94143cea (which contains the contents of #551) should be dropped.

Whilst making this PR, some issued with the `Presentation` class were encountered. Rather than amending them in #551 and continually messing with a PR that is ready for review, they have been fixed here.

Joseph-Edwards:
It seems that some of the benchmarks are failing because of the new SFINAE stuff that prevents, for example, a `Presentation<word_type>` from being trivially converted to a `Presentation<word_type>`. Are these benchmarks up-to-date, and thus worth fixing @james-d-mitchell?

include/libsemigroups/to-presentation.hpp
```diff
@@ -43,173 +47,207 @@ namespace libsemigroups {
...
-    presentation::normalize_alphabet(result);
-    result.alphabet(2 * result.alphabet().size());
-    auto invs = result.alphabet();
-    std::rotate(invs.begin(), invs.begin() + invs.size() / 2, invs.end());
-    result.inverses(std::move(invs));
-    // result.validate();
-    return result;
+  to_inverse_presentation(InversePresentation<Word> const& ip) {
+    return ip;
   }
```
Joseph-Edwards, line 208: As the comment says, is this the best way to prevent an `InversePresentation` from being converted to an `InversePresentation` as if it were a normal `Presentation`? I.e. prevent the addition of extra generators to create inverses from.

Joseph-Edwards: The code snippet above suggests that the old functionality has been removed. It hasn't; that function still exists. The above is an overload that specialises for the `InversePresentation` subclass of `Presentation`.

include/libsemigroups/to-presentation.hpp
```diff
@@ -1,6 +1,6 @@
 //
 // libsemigroups - C++ library for semigroups and monoids
-// Copyright (C) 2023 James D. Mitchell
+// Copyright (C) 2023 - 2024 James D. Mitchell
```
james-d-mitchell, line 3:
```suggestion
// Copyright (C) 2023-2024 James D. Mitchell
```

include/libsemigroups/to-presentation.hpp
```diff
@@ -43,193 +47,207 @@ namespace libsemigroups {
...
+  //! \throws LibsemigroupsException if `p.validate()` throws.
   template <typename Word>
   InversePresentation<Word>
-  to_presentation(InversePresentation<Word> const& p) {
-    return p;
-  }
+  to_inverse_presentation(Presentation<Word> const& p);
 
+  //! No doc
+  // TODO(now) is this the correct way to prevent the nonsensical conversion of
```
james-d-mitchell, line 199: Looks good just add the doc please.

include/libsemigroups/to-presentation.tpp
```diff
@@ -0,0 +1,125 @@
+//
+// libsemigroups - C++ library for semigroups and monoids
+// Copyright (C) 2023 - 2024 James D. Mitchell
```
james-d-mitchell, line 3:
```suggestion
// Copyright (C) 2023-2024 James D. Mitchell
```

tests/test-presentation.cpp
```diff
@@ -19,17 +19,21 @@
...
-#include <unordered_map>  // for operator==, operator!=
-#include <unordered_set>  // for unordered_set
-#include <utility>        // for move, make_pair, swap
-#include <vector>         // for vector, operator==, swap
+#include <algorithm>         // for all_of, equal, fill, sort
+#include <chrono>            // for milliseconds
+#include <cmath>             // for pow
+#include <cstddef>           // for size_t
+#include <cstdint>           // for uint16_t, uint8_t
+#include <ctype.h>           // for isprint
```
james-d-mitchell, line 27:
```suggestion
#include <cctype>           // for isprint
```

tests/test-presentation.cpp
```diff
@@ -19,17 +19,21 @@
 #include <type_traits>
```
james-d-mitchell, line 19: Why is this here and not below?

tests/test-to-presentation.cpp
```diff
@@ -18,8 +18,11 @@
 
 #include <algorithm>         // for fill,  max_element
 #include <cstdint>           // for uint8_t, uint16_t
+#include <ctype.h>           // for isprint
```
james-d-mitchell, line 21:
```suggestion
#include <cctype>           // for isprint
```

include/libsemigroups/to-presentation.hpp
```diff
@@ -43,193 +47,207 @@ namespace libsemigroups {
...
-    result.alphabet_from_rules();
-    result.contains_empty_word(p.contains_empty_word());
-    presentation::normalize_alphabet(result);
-    result.alphabet(2 * result.alphabet().size());
-    auto invs = result.alphabet();
-    std::rotate(invs.begin(), invs.begin() + invs.size() / 2, invs.end());
-    result.inverses_no_checks(std::move(invs));
-    // result.validate();
-    return result;
+  to_inverse_presentation(InversePresentation<Word> const& ip) {
```
james-d-mitchell, line 206: As discussed this should have a `static_assert` instead of returning a copy for consistency with `to_presentation<std::string, std::string>` which won't compile.

Joseph-Edwards: @james-d-mitchell I've tried adding `static_assert(false, "The parameter must not be an InversePresentation");` but that causes a compile time error even if that specific overload never gets called. Am I missing something?

james-d-mitchell: Maybe add something to effect of 
~~~C++
if constexpr (IsInversePresentation<decltype(std::decay_t<x>)>) {
static_assert(false ..)
}
~~~
to the other version of the function, or just delete this function.

Joseph-Edwards: The `IsInversePresentation` stuff didn't work. Unless there's a way to know that the parameter was an `InversePresentation` before it was cast to a `Presentation` in the function call, I can't see how the function would be able to distinguish between the two.

Explicitly deleting the function with `= delete` did seem to work. Thanks!

Joseph-Edwards:
> It seems that some of the benchmarks are failing because of the new SFINAE stuff that prevents, for example, a `Presentation<word_type>` from being trivially converted to a `Presentation<word_type>`. Are these benchmarks up-to-date, and thus worth fixing @james-d-mitchell?

Unless the benchmarks need to pass, I am happy to merge this when you are @james-d-mitchell.

PR #556, opened 2024-07-12
Add Constants

Joseph-Edwards:
This PR finishes of the doc for constants, thus making this file ready for v3.


james-d-mitchell:
Note to self: the documentation doesn't build properly with doxygen v1.11.0 unfortunately. This is probably not an "us" issue, but again the ci job for documentation "passes" when it should really fail.

PR #559, opened 2024-07-29
ci: fix macosx job

james-d-mitchell:


james-d-mitchell:
I've also temporarily, removed the `doc` ci job, since doxygen 1.11.0 seems to be rather broken at the moment.

PR #560, opened 2024-07-29
Account for non-determinism

### LINKED ISSUE #554: Bug in ``todd_coxeter::is_non_trivial``
### Issue
It  seems like there is a bug in `todd_coxeter::is_non_trivial` when debugging is enabled. I tried compiling with both clang++ and g++, and got the error with both.
### Output
```
ToddCoxeter 082: Holt 3
tests/test-todd-coxeter.cpp:3107

tests/test-todd-coxeter.cpp:3122: FAILED:
  REQUIRE( todd_coxeter::is_non_trivial(tc) == tril::TRUE )
with expansion:
  unknown == TRUE

............
```
### How to reproduce
```
./configure --enable-debug
make test_todd_coxeter
./test_todd_coxeter [082]
```

Joseph-Edwards:
Resolves #554.

### DIFF
--- a/tests/test-todd-coxeter.cpp
+++ b/tests/test-todd-coxeter.cpp
@@ -3119,10 +3119,7 @@ namespace libsemigroups {
     presentation::add_rule(p, "cABcabc", "");
 
     ToddCoxeter tc(twosided, p);
-    auto        result = todd_coxeter::is_non_trivial(tc);
-    if (result != tril::TRUE) {
-      REQUIRE(result == tril::unknown);
-    }
+    REQUIRE(todd_coxeter::is_non_trivial(tc) != tril::FALSE);
 
     section_hlt(tc);
     section_felsch(tc);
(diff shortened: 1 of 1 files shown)


james-d-mitchell (commented):
Wouldn't a shorter way of doing this be just `REQUIRE(todd_coxeter::is_non_trivial(tc) != tril::FALSE)`? And thinking about this more, does `is_non_trivial` ever return anything other than TRUE and unknown? Maybe this isn't such a good test :(

Joseph-Edwards:
> Wouldn't a shorter way of doing this be just `REQUIRE(todd_coxeter::is_non_trivial(tc) != tril::FALSE)`?

Yeah, you're right.


> And thinking about this more, does `is_non_trivial` ever return anything other than TRUE and unknown? 

It would return `tril::FALSE` if the `ToddCoxeter` object had one not-obviously-infinite class right?

PR #573, opened 2024-08-02
Add new presentations for transformation monoids

MTWhyte:
This PR adds presentations for the symmetric inverse, full transformation and partial transformation monoids from _[Short presentations for transformation monoids](https://arxiv.org/abs/2406.19294)_.

This is currently a draft, and there are a few things left to do, including:

- Fixing the line break where the default author value is mentioned in the `symmetric_inverse_monoid` documentation. Possibly this has to do with the non-documented comment in the midst of this.

- Adding the presentations for lower degrees -- for mathematical (or in some instances, implementational) reasons, this will need to be done case-by-case.

- Change the default author values for each monoid to use these presentations (seems reasonable!)

- Implement the even-degree-only presentation for the full transformation monoid.

### DIFF
--- a/include/libsemigroups/fpsemi-examples.hpp
+++ b/include/libsemigroups/fpsemi-examples.hpp
@@ -59,9 +59,11 @@ namespace libsemigroups {
       Kassabov   = 16'384,
       Lubotzky   = 32'768,
       Miller     = 65'536,
-      Moore      = 131'072,
-      Moser      = 262'144,
-      Sutov      = 524'288,
+      Mitchell   = 131'072,
+      Moore      = 262'144,
+      Moser      = 524'288,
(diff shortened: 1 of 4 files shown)


MTWhyte:
Done the stuff above. Need to figure out the formatting, I think it's just a version mismatch

MTWhyte:
@james-d-mitchell modulo some details that either I'll figure out how to fix, or perhaps aren't even problems, I think this is worth looking at now

MTWhyte:
N.b. there's a place with a vector called `gens` where there's a very slight duplication of code, but I (at least) feel it's easier to read the way it is

james-d-mitchell (approved):
Looks great thanks @MTWhyte

PR #574, opened 2024-08-02
Fix dot in AhoCorasick

### LINKED ISSUE #570: `dot(AhoCorasick&)` doesn't work
In the proto python bindings for `Dot` I get:

~~~
In [5]:     ac = AhoCorasick()
   ...:     aho_corasick.add_word(ac, [0, 1, 0, 1])
   ...:     aho_corasick.add_word(ac, [0, 1, 1, 0])
   ...:     aho_corasick.add_word(ac, [0, 1, 1, 0, 1])
   ...:     aho_corasick.add_word(ac, [0, 1, 1, 0, 0])
   ...:     aho_corasick.rm_word(ac, [0, 1, 1, 0, 1])
Out[5]: 7

In [6]: ac
Out[6]: <AhoCorasick with 8 nodes>

In [7]: from _libsemigroups_pybind11 import dot

In [8]: dot(ac)
---------------------------------------------------------------------------
LibsemigroupsError                        Traceback (most recent call last)
Cell In[8], line 1
----> 1 dot(ac)

LibsemigroupsError: there is no node named "8"!
~~~

After discussing this with @Joseph-Edwards it looks like the `dot` function is not working properly because the active nodes in the trie are not being looped over, rather just the number of such.

Joseph-Edwards:
Fixes #570, and adds some iterator methods in the process.

### DIFF
--- a/include/libsemigroups/aho-corasick.hpp
+++ b/include/libsemigroups/aho-corasick.hpp
@@ -55,6 +55,8 @@
 //! class in `libsemigroups`.
 namespace libsemigroups {
 
+  class Dot;  // forward decl
+
   //! \ingroup aho_corasick_group
   //!
   //! \brief For an implementation of the Aho-Corasick algorithm
@@ -727,16 +729,6 @@ namespace libsemigroups {
   //! \no_libsemigroups_except
   std::string to_string(AhoCorasick& ac);
 
-  class Dot;  // forward decl
-
-  //! \ingroup aho_corasick_group
-  //!
-  //! \brief Construct a dot object of \p ac.
-  //!
-  //! Construct a \ref Dot object representing the trie of \p ac with suffix
-  //! links.
-  [[nodiscard]] Dot dot(AhoCorasick& ac);
-
   //! \ingroup aho_corasick_group
   //!
   //! \brief Namespace for AhoCorasick helper functions.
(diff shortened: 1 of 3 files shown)


include/libsemigroups/aho-corasick.hpp
```diff
@@ -221,6 +224,89 @@ namespace libsemigroups {
       return _active_nodes_index.size();
     }
 
+    //! \brief Return a the active nodes.
```
james-d-mitchell, line 227:
```suggestion
    //! \brief Return the active nodes.
```

james-d-mitchell (commented):
Looks good @Joseph-Edwards, other than that one typo. 

Shouldn't the doc for the AhoCorasick class say something about what "active nodes" means in this context, and why it might not be the numbers `0, ..., number_of_nodes() - 1`? Thinking a little bit more should we perhaps drop the prefix "active" here? Not sure about this last point.

Joseph-Edwards:
There is currently this in the doc for `validate_active_node_index`: 
> The implementation of [AhoCorasick](file:///home/joseph/Dev/Comp_Maths/libsemigroups/docs/html/classlibsemigroups_1_1_aho_corasick.html) uses two different types of node; active and inactive. An active node is a node that is currently a node in the trie. An inactive node is a node that used to be part of the trie, but has since been removed. It may later become active again after being reinitialised, and exists as a way of minimising how frequently memory needs to be allocated and deallocated for nodes.

Do you think this is worth putting at the top of the file instead?

As for removing the active prefix, there some functions where you can refer to inactive nodes (e.g. `signature_no_checks1. I'm not quite sure why anyone would want to do this, but it gives a sensible output so I don't see any reason to get rid of it, other than maybe it reveals too much unnecessary detail about the implementation.

james-d-mitchell:
@Joseph-Edwards as discussed please move the description of active nodes to the class description, and please ping me when you have. Thanks!

Joseph-Edwards:
> @Joseph-Edwards as discussed please move the description of active nodes to the class description, and please ping me when you have. Thanks!

That's now been pushed @james-d-mitchell

PR #583, opened 2024-08-07
update Konieczny for v3

flsmith:


include/libsemigroups/konieczny.hpp
```diff
@@ -385,6 +385,18 @@ namespace libsemigroups {
       _rho_orb.cache_scc_multipliers(true);
     }
 
+    //! Initialise a Konieczny instance to its default state (or an
+    //! indistinguishable one).
+    //!
+    //! \parameters
+    //! (None)
```
Joseph-Edwards, line 392:
```suggestion
```
For v3, I think we're opting to not including `\parameters (None)` and `\returns (None)`.

tmp.cpp
Joseph-Edwards, line 1: I think this file was added by mistake.

james-d-mitchell:
@flsmith I tried to fix the mem leak indicated in the CI in the commit [b389919](https://github.com/libsemigroups/libsemigroups/pull/583/commits/b389919eef426a8f9f8392a6e23cf6e8a41103b0), I'm not very convinced with my fix though, so can you please double check this for me? 

Also I'm sure there are other places where `_gens` should be `clear`ed as well as deleting each entry, maybe I missed something.

flsmith:
@james-d-mitchell 

> @flsmith I tried to fix the mem leak indicated in the CI in the commit [b389919](https://github.com/libsemigroups/libsemigroups/pull/583/commits/b389919eef426a8f9f8392a6e23cf6e8a41103b0), I'm not very convinced with my fix though, so can you please double check this for me?

This fix looks correct to me. The whole setup of relying on the _one to be the final thing in _gens feels a bit hacky and I might look at it again, but as it is this seems correct.

> Also I'm sure there are other places where `_gens` should be `clear`ed as well as deleting each entry, maybe I missed something.

I took a look and didn't see any. Maybe InternalVecFree should just also clear the vector for simplicity.

james-d-mitchell:
> I took a look and didn't see any. Maybe InternalVecFree should just also clear the vector for simplicity.

Yes, I thought about that but wasn't sure if it was used somewhere else for something else. I'm going to finish modernising the tests, then I think we can merge this, unless there's anything else you are planning on doing @flsmith ?

flsmith:
Not right now, so I think merge @james-d-mitchell

james-d-mitchell:
Currently doing a very slow git bisect using the CI runners, DO NOT MERGE!

PR #581, opened 2024-08-07
Update paths.hpp for v1/3

james-d-mitchell:

PR #587, opened 2024-08-09
paths: add to_human_readble_repr for Reversible

james-d-mitchell:
Forgot some things for ReversiblePaths in the paths-hpp PR #581.


james-d-mitchell:
Expected that both GAP and pybind11 tests will fail

Joseph-Edwards (approved):

PR #588, opened 2024-08-10
Update ranges.hpp for v1/3

james-d-mitchell:
Should be good to go!

### DIFF
--- a/include/libsemigroups/ranges.hpp
+++ b/include/libsemigroups/ranges.hpp
@@ -122,9 +122,9 @@ namespace libsemigroups {
       using output_type = typename InputRange::output_type;
 
       // TODO(1) this is probably not correct, should depend on InputRange
-      static constexpr bool is_finite = true;
+      static constexpr bool is_finite = rx::is_finite_v<InputRange>;
       // TODO(1) this is probably not correct, should depend on InputRange
-      static constexpr bool is_idempotent = true;
+      static constexpr bool is_idempotent = rx::is_idempotent_v<InputRange>;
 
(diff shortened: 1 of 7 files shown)


include/libsemigroups/ranges.hpp
```diff
@@ -38,28 +38,101 @@
...
+    Random& operator=(Random&&) = default;
+
     template <typename InputRange>
     struct Range {
       using output_type = typename InputRange::output_type;
 
-      static constexpr bool is_finite     = true;
-      static constexpr bool is_idempotent = false;
+      // TODO(1) this is probably not correct, should depend on InputRange
+      static constexpr bool is_finite = true;
```
Joseph-Edwards, line 125:
```suggestion
      static constexpr bool is_finite = rx::is_finite_v<InputRange>;
```
Something like this?

Joseph-Edwards (commented):
All looks good to me! I've left a speculative suggestion about one of the TODOs below, but if that's not a good suggestion then this seems good to merge

include/libsemigroups/ranges.hpp
```diff
@@ -38,28 +38,101 @@
...
+      // TODO(1) this is probably not correct, should depend on InputRange
+      static constexpr bool is_idempotent = true;
```
james-d-mitchell, line 127:
```suggestion
      static constexpr bool is_idempotent = rx::is_idempotent_v<InputRange>;
```

james-d-mitchell:
Let's see if that works :)

PR #593, opened 2024-09-06
Update order.hpp for v3

Joseph-Edwards:
This PR updates `order.hpp` for the upcoming release of v3. All of the changes pertain to the doc.


Joseph-Edwards:
Still need to do the Python bindings, but opened this now for visibility.

Issue #597, opened 2024-09-10
Warnings when compiling with `g++ (Ubuntu 13.2.0-23ubuntu4)`


Joseph-Edwards:
The following results in lots of warnings when using `g++ (Ubuntu 13.2.0-23ubuntu4)`:
``` bash
./configure CXX="g++" CXX_FLAGS="-OO"
make test_all
```

The full output can be found here: [make_warnings.txt](https://github.com/user-attachments/files/16948073/new_gcc_warnings.txt), and the breakdown of the warnings is:

|warning|count|
|------------|---------|
|maybe-uninitialized |26 |
|stringop-overflow|  49 |
|uninitialized |   28|
|unused-local-typedefs| 1 |
| unused-parameter | 1|

Joseph-Edwards:
Many of these emanate from `/usr/include/c++/`, so for those I think the most sensible fix for now is to add some `#pragma GCC diagnostic ignored`. I'll have a closer look at the ones coming from `libsemigroups`.

Joseph-Edwards:
I am getting some more compiler warnings, which are being replicated in the Build test_all . . . job of the [gcc workflow](https://github.com/libsemigroups/libsemigroups/actions/runs/12998204712/job/36250875899?pr=632). I'll look into this, and probably consult with @james-d-mitchell. I suspect the solution will be similar to the ones merged in #598.

PR #598, opened 2024-09-11
Fix some warnings

Joseph-Edwards:
This PR removes some compiler warnings introduced by `g++ 13.2`, is highlighted in #597.

### DIFF
--- a/include/libsemigroups/detail/containers.hpp
+++ b/include/libsemigroups/detail/containers.hpp
@@ -866,8 +866,18 @@ namespace libsemigroups {
 
       // Not noexcept because std::array::operator[] isn't
       void push_back(T x) {
+        // The below assertions exist to insure that we are not badly assigning
+        // values. The subsequent pragmas exist to suppress the false-positive
+        // warnings produced by g++ 13.2.0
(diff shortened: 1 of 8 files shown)


Joseph-Edwards:
This is a draft at the moment, because I still have some warnings. Specifically, there are still many instances of `-Wstringop-overflow` both in `libsemigroups/` and `/usr/local/include/c++`. However, a quick google reveals that `gcc` is known to throw false positives of this type, and I suspect that is the case here. For example:

https://github.com/libsemigroups/libsemigroups/blob/0f18cfd263f327309e13102c544f48eba96b8580/include/libsemigroups/detail/containers.hpp#L847-L852

`gcc` gives the warning `writing 1 byte into a region of size 0 [-Wstringop-overflow=]` for line 850 above, but that doesn't seem correct to me (thanks specifically to the assertion in line 849).

I'm not sure what the best way to proceed is, since these warnings crop up in quite a few different places.

Joseph-Edwards:
After some discussions with @james-d-mitchell, I have added a couple of assertions to check that we aren't doing any bad allocations, and then subsequently disabled the warnings deemed to be false-positives.

james-d-mitchell (approved):

PR #600, opened 2024-09-12
Update pbr.hpp for v3

Joseph-Edwards:
This PR updates `pbr.hpp` in preparation for the release of v3.

### DIFF
--- a/include/libsemigroups/pbr.hpp
+++ b/include/libsemigroups/pbr.hpp
@@ -419,7 +419,7 @@ namespace libsemigroups {
     //!
     //! \param x the PBR to validate.
     //!
-    //! \throws LibsemigroupsException if any of the following throw:
+    //! \throws LibsemigroupsException if any of the following occur:
     //! * \p x does not describe a binary relation on an even number of points;
     //! * \p x has a point related to a point that is greater than degree();
     //! * a list of points related to a point is not sorted.
--- a/include/libsemigroups/pbr.hpp
+++ b/include/libsemigroups/pbr.hpp
@@ -31,14 +31,21 @@
 #include "adapters.hpp"  // for Hash
 
 namespace libsemigroups {
+  // The below forward declarations are done so that pbr::validate can be a
+  // friend of PBR. This is required so that validate can access the size of the
+  // private member _vector. An alternative would be to remove these forward
+  // decls and friends, and instead make a PBR member function that returns the
+  // required size.
+  class PBR;
+  namespace pbr {
+    void validate(PBR const& x);
+  }  // namespace pbr
 
-  //! Class for partitioned binary relations (PBR).
-  //!
   //! Partitioned binary relations (PBRs) are a generalisation of bipartitions,
   //! which were introduced by
   //! [Martin and Mazorchuk](https://arxiv.org/abs/1102.0862).
   class PBR {
-    friend void validate(PBR const& x);
+    friend void pbr::validate(PBR const& x);
 
    public:
     //! Type of constructor argument.
@@ -114,7 +121,7 @@ namespace libsemigroups {
     //! performed.
     //!
     //! \sa libsemigroups::validate(PBR const&) and
-    //! libsemigroups::to_pbr(initializer_list_type<int32_t>,
+    //! libsemigroups::to_pbr_no_checks(initializer_list_type<int32_t>,
     //! initializer_list_type<int32_t>)
     PBR(initializer_list_type<int32_t> left,
         initializer_list_type<int32_t> right);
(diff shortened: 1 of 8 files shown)


Joseph-Edwards:
One design choice I have made that I'm not sure about relates to `PBR::one`.

In the old implementation, there was `PBR::one()` and `PBR::one(size_t n)` for creating the identity with the same degree as `this` and with degree `n` respectively. In the new implementation, there is `PBR::one()` and `pbr::one(size_t n)` that do the same thing as the old implementation.

A third option is to do what is done `bipart.hpp`: have `static Bipartition one(size_t n)` and `bipartition one(Bipartition const& f)`. This is seemingly the opposite of what we have at the moment, and I'm not really sure which is better.

tests/test-pbr.cpp
```diff
@@ -171,37 +173,41 @@ namespace libsemigroups {
...
+    REQUIRE_NOTHROW(
+        to_pbr_no_checks({{-3, -1}, {-3, -2, -1, 1, 2, 3}, {-3, -2, -1, 1, 3}},
+                         {{-3, -1, 1, 2, 3}, {-3, 1, 3}, {-3, -2, -1, 2, 3}}));
 
     REQUIRE_NOTHROW(PBR({{}, {}}));
 
     REQUIRE_THROWS_AS(
-        PBR::make({{-4, -1}, {-3, -2, -1, 1, 2, 3}, {-3, -2, -1, 1, 3}},
-                  {{-3, -1, 1, 2, 3}, {-3, 1, 3}, {-3, -2, -1, 2, 3}}),
+        to_pbr_no_checks({{-4, -1}, {-3, -2, -1, 1, 2, 3}, {-3, -2, -1, 1, 3}},
```
james-d-mitchell, line 191: Why is this called no checks but appears to throw here?

include/libsemigroups/pbr.hpp
```diff
@@ -314,56 +314,166 @@ namespace libsemigroups {
...
-  //! Linear in the PBR::degree of \p x.
-  void validate(PBR const& x);
+  //! \sa
+  //! \ref libsemigroups::PBR().
+  //!
+  //! \warning
+  //! No checks are performed on the validity of \p args prior to the
+  //! construction of the PBR object.
+  template <typename... T>
+  inline PBR to_pbr_no_checks(T... args) {
```
james-d-mitchell, line 401: Should be just to_pbr

include/libsemigroups/pbr.hpp
```diff
@@ -314,56 +314,166 @@ namespace libsemigroups {
...
+    // TODO(later) validate_args
+    PBR result(std::forward<T>(args)...);
+    pbr::validate(result);
```
james-d-mitchell, line 404: Can we rename validate to throw_if_something?

james-d-mitchell (changes requested):

tests/test-pbr.cpp
Joseph-Edwards: There are no checks on the arguments before the `PBR` is constructed, so it's possible for bad things to happen before it can be validated. That said, I didn't thoroughly look into what happens during construction, so if nothing bad can happen at that stage, then the `_no_checks` should probably be dropped.

include/libsemigroups/pbr.hpp
Joseph-Edwards: Is that the case for the other `to_pbr` functions, or just the template ones? I was getting linking errors, so I added the inlines but I forgot templates behave differently.

include/libsemigroups/pbr.hpp
Joseph-Edwards: Yeah good shout

Joseph-Edwards:
Once the python bindings are done, and we decided on a better `to_human_readable_repr` than `<PBR of degree n>`, I believe this is good to go.

@james-d-mitchell, do you have any ideas of what else could go in the  `to_human_readable_repr`?

Joseph-Edwards:
I've also noticed that `product_inplace` has become `no_checks`, but we haven't implemented a "checks" version. Is this something we should do? If it's simply a case of checking the degrees are the same then this is easy enough, but if it's something more involved then perhaps it should wait.

james-d-mitchell:
> I've also noticed that `product_inplace` has become `no_checks`, but we haven't implemented a "checks" version. Is this something we should do? If it's simply a case of checking the degrees are the same then this is easy enough, but if it's something more involved then perhaps it should wait.

Not just now, we can live without the checks version for the time being

james-d-mitchell (approved):
Good, thanks @Joseph-Edwards

PR #609, opened 2024-09-23
Froidure pin hpp

james-d-mitchell:


james-d-mitchell:
I renamed the branch to `froidure-pin-base-hpp` because that file's refactor is complete and so we can merge this if the CI passes, and @Joseph-Edwards you don't have any objections.

PR #624, opened 2024-12-11
Add sanitisers to the CI

Joseph-Edwards:
This PR adds the following sanitisers to the CI:
- [x] [AddressSanitizer](https://clang.llvm.org/docs/AddressSanitizer.html)
- [x] [ThreadSanitizer](https://clang.llvm.org/docs/ThreadSanitizer.html)
- [x] [UndefinedBehaviorSanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html)

### DIFF
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -298,8 +298,8 @@ jobs:
          - name: "Run tests with ${{ matrix.type }}Sanitizer . . ."
            run: ${{ matrix.environment_variables }} ./test_all "[quick],[standard]"
          - name: "Check log files"
-           if: ${{ matrix.type != "UndefinedBehaviour" }}
-           run: [ -f ubsan_test.log* ];
+           if: ${{ matrix.type == 'UndefinedBehaviour' }}
+           run: ${{ '[ -f ubsan_test.log* ];' }}
    valgrind:
       timeout-minutes: 60
       runs-on: ubuntu-latest
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -277,7 +277,7 @@ jobs:
             - type: "Address"
               cxx_flags: ""
               ld_flags: ""
-            - type: "UndefinedBehavior"
+            - type: "UndefinedBehaviour"
               cxx_flags: ""
               ld_flags: ""
       steps:
@@ -288,7 +288,7 @@ jobs:
              update_packager_index: false
              install_ccache: true
          - name: "Install dependencies . . ."
-           run: wget -qO- https://apt.llvm.org/llvm.sh | bash -s -- 18
+           run: sudo wget -qO- https://apt.llvm.org/llvm.sh | bash -s -- 18
          - name: "Configure . . ."
            run: |
                 mkdir -p m4 && ./autogen.sh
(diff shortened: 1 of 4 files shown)


Joseph-Edwards:
This will be a draft until all sanitisers are added, and the CI behaves how we would expect it to.

Joseph-Edwards:
The CI is now behaving in a way consistent with my local tests. There is a slight oddity with the fact that, locally, I must specify `LDFLAGS="-fsanitize=..."` to avoid linking errors; conversely, on the CI I must *not* specify `LDFLAGS` to avoid configuration errors. If anyone can explain this I would be grateful, but it seems to work so it's perhaps not worth looking into.

Slightly frustratingly, the sanitisers all seem to have different default behaviour upon detecting an error:
- AddressSanitizer reports the error and immediately terminates the process with a non-zero exit code.
- ThreadSanitizer reports the error to `stderr`, continues running the current process, then returns a non-zero exit code.
- UndefinedBehaviorSanitizer reports the error to `stderr`, continues running the current process, then returns the exit code of the original process.

With some additional flags, ubsan can be made to behave like asan in this regard which seems preferable for our use case. By logging and `grep`-ing the output, we could make ubsan behave like tsan, which I think is the best suited to our CI needs but I'm not sure of that. The documentation of asan seems to suggest that it cannot be made to report errors like tsan.

## Questions
Before I'm ready to un-draft-ify this PR, I have a few questions, some relating to the above discussion.

1. Are we happy with the way errors are reported as discussed above?
2. Should `fail-fast` be set to `false` in the job strategy, so failing jobs don't cancel the other sanitisers? Currently it is `true` by default.
3. Are there any other sanitisers to add? There is documentation for:
  a. [MemorySanitizer](https://clang.llvm.org/docs/MemorySanitizer.html) (probably not, since we have `valgrind`), and
  b. [DataFlowSanitizer](https://clang.llvm.org/docs/DataFlowSanitizer.html).

@james-d-mitchell, these questions are probably best suited for you to answer, but input from anyone else is welcome.

Joseph-Edwards:
Another brief note is that this may need to be changed when `ubuntu-latest` in the workflow points at Ubuntu version 24 rather than version 22

Joseph-Edwards:
I have made some changes to address the questions I had above as a result of a conversation with @james-d-mitchell. Specifically:
 - ubsan now runs all of the tests before checking for warnings;
 - The `fail-fast` strategy of the matrix is now set to `false`, so a failure of one sanitizer will not cause any of the others to be cancelled; and
 - tests marked with the tag `[no-sanitize-undefined]`, `[no-sanitize-address]` or `[no-sanitize-thread]` will not get run in their respective jobs.

The thread sanitizer still fails, and all of these failures seem to come from `Sims` stuff. I think they all relate to lock order inversion, but I'm not certain. For the sake of minimising merge conflicts, I think it is best to address this in #625.

Joseph-Edwards:
In [799014d](https://github.com/libsemigroups/libsemigroups/pull/624/commits/799014db42b784ba5bd8c61e59b3a7f8a947d4ca), a file called `tsan-suppression.cfg` is added. Here, you can specify warnings to be suppressed. Presently, this suppresses deadlock warnings coming from `SimsBase<Sims1or2>::thread_iterator::steal_from` since we suspect that these are false positives. The alternative would be to specify `[no-sanitize-thread]` in all of the tests where this is an issue.

james-d-mitchell (approved):
Looks great. I'm happy with the resolution to the thread sanitizer warnings. @reiniscirpons I think wrote a detailed explanation of why these warnings can be ignored in #625. Maybe we can add that to the `tsan-suppression.cfg` file, when we merge the #625.

PR #626, opened 2024-12-13
Cong hpp

james-d-mitchell:


include/libsemigroups/aho-corasick.tpp
```diff
@@ -97,8 +97,8 @@ namespace libsemigroups {
   }
 
   template <typename Iterator>
-  [[nodiscard]] AhoCorasick::index_type
-  AhoCorasick::traverse_trie(Iterator first, Iterator last) const {
+  AhoCorasick::index_type AhoCorasick::traverse_trie(Iterator first,
```
Joseph-Edwards, line 100: What are the reasons for removing `[[nodiscard]]` here?

james-d-mitchell: Yeah, it should only be used with the function declaration in the hpp file

Joseph-Edwards (approved):
The changes here all seem reasonable to me. The tests pass on my machine, and compilation doesn't spit out any more warnings than before. I haven't interrogated the doc fully, but @james-d-mitchell and I decided to do that in a new PR (which I will begin work on imminently). Thanks for this behemoth PR @james-d-mitchell!

PR #629, opened 2025-01-06
Replace `Transf::make` with `to<Transf>`, and same for `Perm` and `PPerm`

Joseph-Edwards:
This PR replaces the static member function `make` from various `PTransf` classes with a free function template `to<Class>`. This came after discussions with @james-d-mitchell.

### DIFF
--- a/include/libsemigroups/matrix.hpp
+++ b/include/libsemigroups/matrix.hpp
@@ -7035,7 +7035,7 @@ namespace libsemigroups {
     //! \par Example
     //!
     //! \code
-    //! auto x == ProjMaxPlusMat<>::make({{-2, 2, 0}, {-1, 0, 0}, {1, -3,
+    //! auto x == ProjMaxPlusMat<>::to_matrix({{-2, 2, 0}, {-1, 0, 0}, {1, -3,
     //! 1}}));
     //! // returns {{-1, 0, -1}, {-2, -1, -2}, {-1, 0, -1}}
     //! matrix::pow(x, 100);
@@ -7783,7 +7783,7 @@ namespace libsemigroups {
     //! \par Example
     //!
     //! \code
-    //! auto x = BMat<>::make({{1, 0, 0}, {0, 0, 1}, {0, 1, 0}});
+    //! auto x = BMat<>::to_matrix({{1, 0, 0}, {0, 0, 1}, {0, 1, 0}});
     //! matrix::row_space_size(x); // returns 7
     //! \endcode
     template <typename Mat, typename = std::enable_if_t<IsBMat<Mat>>>
--- a/include/libsemigroups/matrix.hpp
+++ b/include/libsemigroups/matrix.hpp
@@ -7035,13 +7035,14 @@ namespace libsemigroups {
     //! \par Example
     //!
     //! \code
-    //! auto x == ProjMaxPlusMat<>::to_matrix({{-2, 2, 0}, {-1, 0, 0}, {1, -3,
+    //! auto x == ProjMaxPlusMat<>::make({{-2, 2, 0}, {-1, 0, 0}, {1, -3,
     //! 1}}));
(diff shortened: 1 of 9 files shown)


include/libsemigroups/transf.hpp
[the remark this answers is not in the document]
Joseph-Edwards: After [3242898](https://github.com/libsemigroups/libsemigroups/pull/629/commits/32428981f2deaec0d31056a806b258f9010699f6), the logic in this comment doesn't hold anymore, but the `make` functions still exist. I will have a look and see if I can get rid of them now.

Joseph-Edwards:
Downstream will fail until https://github.com/libsemigroups/libsemigroups_pybind11/pull/212 is merged. Otherwise, this is good to go.

Joseph-Edwards:
One of the tickets on clickup says to use `to_transf`(which would now actually be `to<Transf>`) instead of raw `Transf` calls in the tests. Since most of the cases where this is relevant relates to Sims or congruence stuff, should I hold off on doing that @james-d-mitchell ?

include/libsemigroups/matrix.hpp
```diff
@@ -7035,7 +7035,7 @@ namespace libsemigroups {
     //! \par Example
     //!
     //! \code
-    //! auto x == ProjMaxPlusMat<>::make({{-2, 2, 0}, {-1, 0, 0}, {1, -3,
+    //! auto x == ProjMaxPlusMat<>::to_matrix({{-2, 2, 0}, {-1, 0, 0}, {1, -3,
```
james-d-mitchell, line 7038: Don't think this is (or was) correct, maybe add a TODO or a just fix it?

include/libsemigroups/matrix.hpp
```diff
@@ -7783,7 +7783,7 @@ namespace libsemigroups {
     //! \par Example
     //!
     //! \code
-    //! auto x = BMat<>::make({{1, 0, 0}, {0, 0, 1}, {0, 1, 0}});
+    //! auto x = BMat<>::to_matrix({{1, 0, 0}, {0, 0, 1}, {0, 1, 0}});
```
james-d-mitchell, line 7786: Same comment

include/libsemigroups/transf.hpp
```diff
@@ -50,7 +50,7 @@
 #include "debug.hpp"      // for LIBSEMIGROUPS_ASSERT
 #include "exception.hpp"  // for LIBSEMIGROUPS_EXCEPTION
 #include "hpcombi.hpp"    // for HPCombi::Transf16
-#include "types.hpp"      // for SmallestInteger
+#include "types.hpp"      // for SmallestInteger, enable_if_is_same
```
james-d-mitchell, line 53: Looks like we're not using `enable_if_is_same` any more.

include/libsemigroups/transf.hpp
```diff
@@ -1140,6 +1108,68 @@ namespace libsemigroups {
...
+  //! \tparam Return the return type.
+  //! \tparam OtherContainer universal reference for the type of the container.
+  //!
+  //! \param cont the container.
+  //!
+  //! \returns A \ref Transf instance with degree \c N.
+  //!
+  //! \throw LibsemigroupsException if any of the following hold:
+  //! * the size of \p cont is incompatible with \ref container_type.
+  //! * any value in \p cont exceeds `cont.size()` and is not equal to
```
james-d-mitchell, line 1133: I don't think the "and is not equal to UNDEFINED" should be here.

james-d-mitchell (changes requested):
Some minor tweaks and this should be good to go. Thanks @Joseph-Edwards

james-d-mitchell:
> One of the tickets on clickup says to use `to_transf`(which would now actually be `to<Transf>`) instead of raw `Transf` calls in the tests. Since most of the cases where this is relevant relates to Sims or congruence stuff, should I hold off on doing that @james-d-mitchell ?

Yup better hold off for now.

PR #630, opened 2025-01-06
Fix coverage and downstream CI jobs

Joseph-Edwards:
This PR updates the coverage workflow for use in Ubuntu 24.04, and fixes a bug in the downstream job. 

The coverage update included suppressing some errors produced by `lcov` which are documented below. It's not clear to me whether these errors are caused by us, Catch2, `gcc`, `lcov` or something else, so if anyone could shed some light that would be greatly appreciated. An interesting discussion on a very similar issue can be found at https://github.com/linux-test-project/lcov/issues/341.

For now, suppressing the errors seems to be a suitable temporary solution.

## The errors

```bash
geninfo: ERROR: mismatched end line for _ZN13libsemigroups12_GLOBAL__N_134ns_CATCH2_INTERNAL_TEMPLATE_TEST_131CATCH2_INTERNAL_TEMPLATE_TEST_1IJNS1_8TypeListIJNS_6detail11RewriteTrieEEEENS3_IJNS4_15RewriteFromLeftEEEEEEC2Ev at /home/runner/work/libsemigroups/libsemigroups/tests/test-knuth-bendix-3.cpp:84: 130 -> 84

geninfo: ERROR: Unexpected negative count '-11' for /home/runner/work/libsemigroups/libsemigroups/include/libsemigroups/detail/felsch-graph.tpp:321.
        Perhaps you need to compile with '-fprofile-update=atomic
```

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -55,7 +55,7 @@ jobs:
               sudo apt-get --yes update
               sudo apt-get install python3 --yes
               sudo apt-get install python3-pip --yes
-              sudo -H pip3 install -r requirements.txt
+              pip3 install -r requirements.txt
           - name: "Building libsemigroups_pybind11 . . ."
             run: |
               cd libsemigroups_pybind11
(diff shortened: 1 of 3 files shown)


Joseph-Edwards:
Whilst the coverage job passes, nothing seems to actually get uploaded to the Codecov website. If I understand correctly, this is something we've had trouble with in the past.

PR #631, opened 2025-01-07
Replace some `to_class` functions with `make<Class>`

Joseph-Edwards:
This PR builds on #629 by converting several `to_class` functions to  `make<Class>` functions. Specifically, the functions in question are:

- [x] `to_bipartition`
- [x] `to_blocks`
- [x] `to_forest`
- [x] `to_matrix`
- [x] `to_pbr`
- [x] `to_perm`
- [x] `to_pperm`
- [x] `to_transf`
- [x] `to_word_graph`

If I have missed any, please let me know!

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -70,13 +70,13 @@ Some more conventions:
 
   * A constructor should be used to construct an object without checking its
     arguments.
-  * A `make` function should be a member of a helper namespace (such as
-    `inverse_presentation` or `blocks`) that is used to construct an object
-    from non-libsemigroups objects such as containers, integers and strings.
-    It should check its arguments.
-  * A `to` function should be a specialisation of a function template that is
-    used to convert from one `libsemigroups` type to another. A typical
-    signature might look something like
+  * A `make` function should be a free function template in the libsemigroups
+    namespace that is used to construct an object from non-libsemigroups
+    objects such as containers, integers and strings. It should check its
+    arguments.
+  * A `to` function should a free function template in the libsemigroups
+    namespace that is used to convert from one `libsemigroups` type to another.
+    A typical signature might look something like
     `to<ToddCoxeter>(Presentation<word_type> p)`. It should check its
     arguments.
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -64,10 +64,21 @@ Some more conventions:
   - functions that are (or should be) implemented for all of the public facing
     classes in ``libsemigroups``, such as, for example
     `to_human_readable_repr`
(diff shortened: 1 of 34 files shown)


Joseph-Edwards:
One slight difference with this is relates to templates. Suppose I want to perform some computation on lots of different types of `Transf`, so I make a function template with `TransfType` as the template parameter. Using the old method, I could construct this object like: 
``` cpp
to<TransfType>(...)
```

Now, however, there is no way to replicate this behaviour. Instead, my function template would need `N` and `Scalar` as template parameters. The new function call would look something like:
```
transf::make<N, Scalar>(...)
```

It's not clear to me of one of these is better than the other, but nonetheless a difference that seemed important to point out.

Joseph-Edwards:
I think it's sensible to make the remaining changes in a separate PR, as they might be slightly more involved. I expect downstream to fail until the corresponding `pybind11` stuff gets merged.

james-d-mitchell:
> One slight difference with this is relates to templates. Suppose I want to perform some computation on lots of different types of `Transf`, so I make a function template with `TransfType` as the template parameter. Using the old method, I could construct this object like:
> 
> ```c++
> to<TransfType>(...)
> ```
> 
> Now, however, there is no way to replicate this behaviour. Instead, my function template would need `N` and `Scalar` as template parameters. The new function call would look something like:
> 
> ```
> transf::make<N, Scalar>(...)
> ```
> 
> This will also influence how we write tests.
> 
> It's not clear to me if one of these is better than the other, but nonetheless a difference that seemed important to point out.

Now that you mention it, it probably makes more sense to have consistent syntax, by which I mean I think we’d want `make<Transf<N, Scalar>>(…)`

Joseph-Edwards:
> Now that you mention it, it probably makes more sense to have consistent syntax, by which I mean I think we’d want `make<Transf<N, Scalar>>(…)`

Would this still be in the `transf` namespace? Or have all of the `make` functions in the `libsemigroups` namespace?

james-d-mitchell:
Same as `to` so in the `libsemigroups` namespace

james-d-mitchell:
Good to go?

james-d-mitchell:
What happened to `to_presentation`? Shouldn't that be converted to `to` also?

Joseph-Edwards:
> What happened to `to_presentation`? Shouldn't that be converted to `to` also?

I was going to wait until #626 was merged, and do presentations along with the other to_congruence type functions. I can do it now if that's preferable?

james-d-mitchell:
> > What happened to `to_presentation`? Shouldn't that be converted to `to` also?
> 
> I was going to wait until #626 was merged, and do presentations along with the other to_congruence type functions. I can do it now if that's preferable?

Later is ok! Just wanted to check it didn't get missed out

PR #632, opened 2025-01-08
Update doc for `fpsemi-examples.hpp`

Joseph-Edwards:
This PR builds upon #517 by making the minor tweaks to the doc to make it more consistent with rest of the code base, and to address some doxygen warnings.


Joseph-Edwards:
There are still a few undocumented functions, but they seem to require slightly more specialist knowledge. @james-d-mitchell, would you be able to help with this please, or alternatively make the call to leave them undocumented.

src/fpsemi-examples.cpp
```diff
@@ -954,6 +820,111 @@ namespace libsemigroups {
...
+      // V6
+      presentation::add_rule_no_checks(
+          p, 3113_w + pow(1_w, n - 2), 113_w + pow(1_w, n - 2) + 3_w);
+      // V7
+      presentation::add_rule_no_checks(
+          p, 3112_w + pow(1_w, n - 2), 112_w + pow(1_w, n - 2) + 3_w);
+      return p;
+    }
+
+    Presentation<word_type> partition_monoid_Machine(size_t n) {
```
Joseph-Edwards, line 879: The presentations provided here don't contain the empty rule. Should it?

james-d-mitchell: Yup

james-d-mitchell:
> Codespell is failing because `groupe` appears in the `.bib` file. @james-d-mitchell Do you think it make more sense to skip `.bib` files in Codespell, and `groupe` to the dictionary, or both?

Probably add to the dictionary is best thanks @Joseph-Edwards

include/libsemigroups/fpsemi-examples.hpp
```diff
@@ -22,710 +22,1318 @@
...
-    //!
-    //! \throws LibsemigroupsException if `n < 3`
-    //! \throws LibsemigroupsException if `val` is not `author::FitzGerald`
-    //!
-    //! [10.1017/s0004972700037692]: https://doi.org/10.1017/s0004972700037692
-    [[nodiscard]] Presentation<word_type>
-    uniform_block_bijection_monoid(size_t n, author val = author::FitzGerald);
-
-    //! A presentation for the partition monoid.
+    //! \name Specific presentations
```
Joseph-Edwards, line 40: This (along with `\name Default presentations` further down in the file) constructs a *member group* which splits the documentation for the function definitions into two parts; one for the default functions (e.g. `partition_monoid`), and one for the specific presentations (e.g. `partition_monoid_HR05`). I think this is nice, but for some reason it means the documentation appears in the order it is in the `.hpp` file, rather than alphabetically. Is this okay? Alternatives are reverting to no member groups, sorting the source file, or doing something completely different such as defining new groups.

include/libsemigroups/fpsemi-examples.hpp
```diff
@@ -22,710 +22,1318 @@
...
+    //!
+    //! \param l the size of the claimed monoid.
+    //!
+    //! \returns a value of type `Presentation<word_type>`.
+    //!
+    //! \exceptions
+    //! \no_libsemigroups_except
     [[nodiscard]] Presentation<word_type>
-    renner_type_D_monoid(size_t l, int q, author val = author::Any);
+    not_renner_type_D_monoid_Machine(size_t l);
```
Joseph-Edwards, line 731: After a discussion with @james-d-mitchell, we decided we should delete all `Machine` presentations. I have done so for all of them except this one. It had no reference and, from what I can tell, it was an example that we came up with that has actual mathematical significance. If I understand correctly, it is the Iwahori-Hecke deformation 0 analogue of an incorrect presentation given by Godelle for the Renner type D monoid (Godelle's presentation being of Iwahori-Hecke deformation 1).

It's not really clear to me what we should do with this function. It seems sensible to keep, but it requires a better name if we keep it. I'm very open to suggestions :).

Theis

Joseph-Edwards:
Other than the few questions outlined above, I think this is good to go @james-d-mitchell.

include/libsemigroups/fpsemi-examples.hpp
Joseph-Edwards: Sort them to be alphabetical

include/libsemigroups/fpsemi-examples.hpp
Joseph-Edwards: Keep q as a parameter and give all references possible

Joseph-Edwards:
Move these functions to the presentation namespace.

Joseph-Edwards:
> Move these functions to the presentation namespace.

I moved all of these functions to the `presentation` namespace but I didn't like the result; they felt too did dissimilar to the other functions in that namespace.

All of the examples are now members of the `presentation::examples` namespace. This has the disadvantage that fully qualified names are quite long, but I don't think this is too problematic.

If you're happy with this, then this is good to go @james-d-mitchell .

Issue #614, opened 2024-10-16
Replace double backticks with single backtics in docs


Joseph-Edwards:
Throughout the codebase, we use a mixture of double backticks, single backticks and `\c` commands to indicate that some text should be interpreted as code. The `\c` command feels the most *Doxygenic*, but it only works for one-word code blocks. In Doxygen 1.10 both double backticks and single backticks seem to work fine. However, in later versions, the double backticks cause issues that become resolved when changed to single backticks.

Using Doxygen 1.12 seems to fix the strange bug that makes some links spuriously resolve to `BMat`, so it seems sensible to me to update.

PR #634, opened 2025-01-14
Update doxygen

Joseph-Edwards:
Addresses #614 by replacing double backticks with single backticks. This also updates the `Doxyfile` to be compatible with doxygen 1.13, and pulls the most recent updates from the `doxygen-awesome-css` submodule.


Joseph-Edwards:
@reiniscirpons, I noticed that in `fixup.py` there are a few functions that claim to only work for `Doxygen <= 1.10.0`. Is this something that I need to address?

reiniscirpons:
@Joseph-Edwards , I only put 1.10.0 because that was the highest available version at the time. Since we are relying on parts of Doxygen that are not its public api, any minor or even bugfix version change of Doxygen is potentially breaking for us. If the docs still generate fine and the html output is not messed up/the sidebar works, then potentially the fixup script is still fine for versions above 1.10.0.

I'll take a deeper look into the fixup functions tomorrow. Maybe there is a way of doing what we need directly through doxygen with these newer versions, so we can remove the fixup file altogether.

PR #640, opened 2025-02-09
Cong hpp doc

Joseph-Edwards:
This PR homogenises the doc between the various congruence-like classes:
 - `ToddCoxeter`;
 - `KnuthBendix`;
 - `Kambites`; and
 - `Congruence`.

### DIFF
--- a/docs/Doxyfile
+++ b/docs/Doxyfile
@@ -307,7 +307,9 @@ ALIASES                = "exceptions=\par Exceptions" \
                          "cong_common_throws_if_started=\throws LibsemigroupsException if \ref started returns \c true." \
                          "cong_common_throws_if_letters_out_of_bounds=\throws LibsemigroupsException if any of the values pointed at by the iterators is out of range, i.e. they do not belong to `presentation().alphabet()` and `Presentation::validate_word` throws." \
                          "ref_todd_coxeter=\ref todd_coxeter_class_group \"ToddCoxeter\"" \
-                         "ref_knuth_bendix=\ref knuth_bendix_class_group \"KnuthBendix\""
+                         "ref_knuth_bendix=\ref knuth_bendix_class_group \"KnuthBendix\"" \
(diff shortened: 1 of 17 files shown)


Joseph-Edwards:
It's not clear to me where `to_human_readable_repr` should be documented. At the moment, it is the only function in the top level of the `ToddCoxeter` page. Of all the groups we have defined, it feels most like an accessor, but it's not a member function so that's also confusing. Very open to suggestions.

james-d-mitchell:
> It's not clear to me where `to_human_readable_repr` should be documented. At the moment, it is the only function in the top level of the `ToddCoxeter` page. Of all the groups we have defined, it feels most like an accessor, but it's not a member function so that's also confusing. Very open to suggestions.

I think we've documented the `to_human_readable_repr` at the top level of the other congruence-like classes too, so let's just go with this for consistency

Joseph-Edwards:
I think this is good to go @james-d-mitchell

james-d-mitchell:
Looks generally great! Don't we need to define \ref_kambites so that the references lead to the correct place? Likewise for Congruence?

Joseph-Edwards:
> Looks generally great! Don't we need to define \ref_kambites so that the references lead to the correct place? Likewise for Congruence?

Good spot. It was still making links to the class itself, rather than the new topic page. This is fixed now.

I have two questions relating to helper namespaces before I'm happy to merge.

Firstly, we reference the helper namespaces (not the `thing_helpers_group`) in several places in the doc. If we click those links, it takes us into the void (i.e. takes us to a place not present in the sidebar). Is this okay, or should they be added to the side bar?

Secondly, the `Kambites` and `Congruence` helper namespaces only contain common helpers. Therefore, if we were to have `kambites_heplers_group` and `congruence_helpers_group` (to remain consistent with KB and TC) these would both be empty. Is that okay? Or should we have nothing at all?

With these questions addressed, I'd be happy to merge.

PR #642, opened 2025-02-19
To froidure pin

james-d-mitchell:


include/libsemigroups/to-froidure-pin.hpp
```diff
@@ -20,194 +20,261 @@
 #define LIBSEMIGROUPS_TO_FROIDURE_PIN_HPP_
 
 #include <cstddef>      // for size_t
+#include <memory>       // for unique_ptr
 #include <type_traits>  // for enable_if_t, is_base_of
```
Joseph-Edwards, line 25: Should there be `#include cong-class.hpp` here too? My initial though is yes, as later we call `cong.run()` so a forward decl isn't sufficient (at least this is what I think `pybind11` was complaining about). If yes, the problem is that we now have a circular include:
`to-froidure-pin.hpp` includes `cong-class.hpp` includes `to-todd-coxeter.hpp` includes `to-froidure-pin.hpp`.

As a test, I hackily tried to remove the `to-todd-coxeter.hpp`'s depended on `to-froidure-pin.hpp` which seemed to make `pybind11` happy. Not sure what the right thing to do is.

include/libsemigroups/to-todd-coxeter.hpp
```diff
@@ -15,62 +15,112 @@
...
+  //! \throws LibsemigroupsException if `kb.kind()` is not \ref
+  //! congruence_kind::twosided.
+  //!
+  //! \throws LibsemigroupsException if `kb.number_of_classes()` is not finite.
+  //! Use ToddCoxeter(knd, kb.presentation()) in this case.
+  template <template <typename...> typename Thing,
+            typename Word,
+            typename Rewriter,
+            typename ReductionOrder>
+  auto to(congruence_kind knd, KnuthBendix<Word, Rewriter, ReductionOrder>& kb)
```
Joseph-Edwards, line 119: Is there any reason why this is implemented for `KnuthBendix`, but the `to-froidure-pin` counterpart is implemented for `KnuthBendixImpl`?

james-d-mitchell: Yeah good question, we need the type of the words in the knuthbendix so we can use the same type in Todd-Coxeter

james-d-mitchell:
This is complete from my perspective, unless the CI fails. @Joseph-Edwards

james-d-mitchell:
@Joseph-Edwards needs to be rebased on the current v3

Joseph-Edwards (approved):
I think this is good to go now. Thanks @james-d-mitchell!

PR #644, opened 2025-02-25
ci: move x86-32 to os.yml

james-d-mitchell:


james-d-mitchell:
Seems to me that this just can't work at the moment. The issue appears to be that the github actions runner requires node version 20, but the version of ubuntu in the docker container at 

https://hub.docker.com/r/i386/ubuntu

is so ancient that it doesn't contain node version 20. I'm not sure if there's anywhere else to get i386 from, so if my attempts to install node 20 in the docker container fail, then I give up. I think it might be a good idea to keep this PR open to remind us why this didn't work.

james-d-mitchell:
Seems related to:

https://github.com/actions/runner/issues/2115

PR #653, opened 2025-03-03
Reorganize the CI

Joseph-Edwards:
This PR reorganizes the CI, and extends the things it covers, as outlined below.

### Changes

- [x] Valgrind runs with the sanitizers.
- [x] Config flags tested more thoroughly.
- [x] Brief quick test job added.
- [x] `os.yml` contains OS tests.
- [x] Different compilers tested.

PR #654, opened 2025-03-04
Fix compile warnings

james-d-mitchell:
This PR fixes all the compiler warnings present on my computer (except those arising in `stephen.hpp`, didn't want to collide with what @reiniscirpons is doing) when compiling with both gcc (14) and clang (19), looking at the output of 

https://github.com/libsemigroups/libsemigroups/actions/runs/13659358677/job/38186609587?pr=653

It looks like I haven't got them all. Maybe we should merge #653 first, then get this to fully work if possible.


tests/test-froidure-pin-transf.cpp
```diff
@@ -19,12 +19,13 @@
 #include <algorithm>  // for min
 #include <cstddef>    // for size_t
 #include <cstdint>    // for uint_fast8_t, uint16_t
-#include <iterator>
-#include <vector>  // for vector
+#include <iterator>   // for ??
```
Joseph-Edwards, line 22:
```suggestion
#include <iterator>   // for make_reverse_iterator
```

Joseph-Edwards (commented):
All looks good to me @james-d-mitchell . I've made one suggestion, but this is really minor so we can take take it or leave it.

I still get some compile warnings on my machine, so I think we should hold off merging for now, and see if #653 gives more insight. I'm happy to have a go at fixing some of the warnings that arises for if that'd be helpful

Joseph-Edwards:
> Might be worthwhile not actually running the tests in the warning jobs, since they’re extremely slow

Yeah I suppose there's no point really. I'll open a new PR with this change, and also the arm runner

james-d-mitchell:
> > Might be worthwhile not actually running the tests in the warning jobs, since they’re extremely slow
> 
> Yeah I suppose there's no point really. I'll open a new PR with this change, and also the arm runner

Tomorrow!

PR #663, opened 2025-03-06
More updates to the CI

Joseph-Edwards:
This PR:
 - [x] specifies the compilers we use in the "Compile warnings" workflow;
 - [x] doesn't run the tests in the compiler workflow (because we only care if they compile);
 - [x] adds an Ubuntu-arm runner; 
 - [x] fixes the behaviour of `make lint` to fail if we see a linting error; and
 - [x] moves Catch2 into `third_party` with a wrapper that suppresses some compiler warnings.

This last one should probably be in its own PR, but I got in a bit of a muddle with my branches.


Joseph-Edwards:
I think this is now finally good to go @james-d-mitchell!

PR #667, opened 2025-03-06
Make `add_generating_pair` a void function

Joseph-Edwards:
This PR makes `add_generating_pair` a void function, rather than returning its first element. This makes it consistent with the rest of the helpers that modify their first element.


Joseph-Edwards:
Downstream will fail until https://github.com/libsemigroups/libsemigroups_pybind11/pull/225 is merged.

PR #691, opened 2025-03-15
Even better reporting

james-d-mitchell:
This PR adds some even better reporting than before. This exposed some issues, which are also fixed in this PR (namely with `Congruence` not doing the correct thing in some cases, leading to not being `finished` after `run` has been called, also refactor `Race.run_until` to not use `run_for`. The latter seems to work, let's see what the CI says).

### DIFF
--- a/include/libsemigroups/konieczny.tpp
+++ b/include/libsemigroups/konieczny.tpp
@@ -15,6 +15,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+#include "libsemigroups/detail/report.hpp"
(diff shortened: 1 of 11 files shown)


james-d-mitchell:
I discovered the issue resolved in #690 when doing this, so I haven't yet finished.

include/libsemigroups/konieczny.tpp
```diff
@@ -15,6 +15,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+#include "libsemigroups/detail/report.hpp"
```
james-d-mitchell, line 18:
```suggestion
```

PR #692, opened 2025-03-17
Reclassify tests

Joseph-Edwards:
This PR reclassifies the tests so that we conform with the table specified in [CONTRIBUTING.rst](https://github.com/libsemigroups/libsemigroups/blob/36f00facebb5ce4235af2ab6bc1fae1fab7f5ae9/CONTRIBUTING.rst), and makes minor changes to make the test more consistent.

Specifically:
 - [X] The time-related tag of each test case conforms with: 
   - "quick" < 200ms;
   - 200ms < "standard" < 3s;
   - 3s < "extreme".
 - [X] Reporting is enabled for every "extreme" test;
 - [X] reporting is disabled for every "quick" and "standard" test;
 - [X] "fail" is marked for old "extreme tests" that hadn't finished after several hours; and
 - [X] "fail" is marked for any test that does not pass.

I tried my best to fix any failing tests but, in several of them, it wasn't clear to me what was actually being tested, so these have been marked fail. I welcome any suggestions on if/how these should be fixed. 

With these changes, using an Intel i7-13700H cpu with 64 GB of RAM and configuring using `./configure CXX="ccache g++" CXXFLAGS="-g -O2"`, the:
 - "quick" tests run in ~8 seconds;
 - "standard" tests run in ~1 minute 20 seconds; and
 - "extreme" tests run in ~1 hour 10 minutes.

The output generated by the extreme tests can be found below for reference:
[extreme.log.zip](https://github.com/user-attachments/files/19277450/extreme.log.zip)

### DIFF
--- a/tests/test-todd-coxeter.cpp
+++ b/tests/test-todd-coxeter.cpp
@@ -2747,7 +2747,7 @@ namespace libsemigroups {
   }
 
   // This test seems to segfault, including on debug mode, and it causes JDE's
-  // terminal to crash making it tought to figure out what's going wrong.
+  // terminal to crash making it tough to figure out what's going wrong.
   LIBSEMIGROUPS_TEST_CASE("ToddCoxeter",
                           "068",
                           "Walker 5",
@@ -2793,7 +2793,7 @@ namespace libsemigroups {
   }
 
   // This test seems to segfault, including on debug mode, and it causes JDE's
-  // terminal to crash making it tought to figure out what's going wrong.
+  // terminal to crash making it tough to figure out what's going wrong.
   LIBSEMIGROUPS_TEST_CASE("ToddCoxeter",
                           "069",
                           "not Walker 6",
@@ -3774,7 +3774,7 @@ namespace libsemigroups {
   }
 
   // This test seems to segfault, including on debug mode, and it causes JDE's
-  // terminal to crash making it tought to figure out what's going wrong.
+  // terminal to crash making it tough to figure out what's going wrong.
   LIBSEMIGROUPS_TEST_CASE("ToddCoxeter",
                           "096",
                           "http://brauer.maths.qmul.ac.uk/Atlas/spor/M22",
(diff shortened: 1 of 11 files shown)


tests/test-todd-coxeter.cpp
```diff
@@ -2747,10 +2746,12 @@ namespace libsemigroups {
     REQUIRE(tc.number_of_classes() == 36'412);
   }
 
+  // This test seems to segfault, including on debug mode, and it causes JDE's
+  // terminal to crash making it tought to figure out what's going wrong.
   LIBSEMIGROUPS_TEST_CASE("ToddCoxeter",
                           "068",
                           "Walker 5",
-                          "[todd-coxeter][extreme]") {
+                          "[todd-coxeter][fail]") {
```
Joseph-Edwards, line 2754: This test (and several others that I have marked with the same comment) consistently crashed my terminal. It's not clear if this is a me issue, but I have marked them with fail to be safe. If anyone else is able to reproduce, or not, I'd be interested to know!

Joseph-Edwards:
> Looks good, but it looks like you removed the no-valgrind, tag from some of the tests, which shows in the run time of the valgrind ci job (47 minutes), which I think is higher than it was before. Can you please put the no-valgrind tags back in? The rationale being that these tests don't test anything different than other tests that run in a more reasonable amount of time.

I only removed the `no-valgrind` tags from tests that are no longer "quick", as these don't get run by valgrind anyway. I'm happy to put the `no-valgrind` tags back in for clarity, but I don't think this is why valrgind was running slower.

james-d-mitchell:
> > Looks good, but it looks like you removed the no-valgrind, tag from some of the tests, which shows in the run time of the valgrind ci job (47 minutes), which I think is higher than it was before. Can you please put the no-valgrind tags back in? The rationale being that these tests don't test anything different than other tests that run in a more reasonable amount of time.
> 
> 
> 
> I only removed the `no-valgrind` tags from tests that are no longer "quick", as these don't get run by valgrind anyway. I'm happy to put the `no-valgrind` tags back in for clarity, but I don't think this is why valrgind was running slower.

Right you are, I thought the standard tests were also run. That being the case I'll merge, thanks @Joseph-Edwards !

PR #694, opened 2025-03-17
Add test analysis

Joseph-Edwards:
This PR adjusts the output of the `Quick tests/g++` job so that the results can be uploaded to Codecov. Moving forward, this will allow us to track the runtime of individual tests, and see if there are tests that fail more often than others. This should only introduce a small overhead, since these tests are getting run anyway; now we just keep hold of the output. 

This PR also removes the `--disable-hpcombi` flag from several otherwise default configurations, since I couldn't really see what it was included. If there was a good reason for this, I'm happy to revert this change.

### DIFF
--- a/.github/workflows/compilers.yml
+++ b/.github/workflows/compilers.yml
@@ -50,7 +50,7 @@ jobs:
         run: |
           mkdir -p m4
           ./autogen.sh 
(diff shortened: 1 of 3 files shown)


james-d-mitchell:
Nice thanks @Joseph-Edwards

PR #698, opened 2025-03-20
Rm extra deduction guides

james-d-mitchell:


james-d-mitchell:
@Joseph-Edwards Any idea why the ci is failing?

reiniscirpons:
Looks like downstream fails because of some changes/weird interactions between `setuptools` and `packaging` in pybind11, see e.g. https://github.com/pypa/setuptools/issues/4892 for a similar bugreport from this morning. I think we just need to be more strict in versioning `setuptools` and `packaging` in `libsemigroups_pybind11` .

PR #708, opened 2025-03-25
Move `Paths` helper functions to the `paths` namespace

Joseph-Edwards:
This PR moves the `Paths` helper functions to the `paths` namespace.

### DIFF
--- a/include/libsemigroups/detail/path-iterators.hpp
+++ b/include/libsemigroups/detail/path-iterators.hpp
@@ -191,7 +191,8 @@ namespace libsemigroups {
       }
 
       [[nodiscard]] node_type target() const noexcept {
-        if (_it == cend_pilo(_it.word_graph()) && _length == UNDEFINED) {
+        if (_it == const_pilo_iterator<Node>(&_it.word_graph(), 0, 0, 0)
+            && _length == UNDEFINED) {
           return UNDEFINED;
         }
         return _it.target();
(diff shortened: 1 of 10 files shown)


include/libsemigroups/detail/path-iterators.hpp
```diff
@@ -337,10 +338,11 @@ namespace libsemigroups {
                              size_type                   max)
           : _count(source == UNDEFINED
                        ? 0
-                       : number_of_paths(*ptr, source, target, min, max)),
+                       //  : number_of_paths(*ptr, source, target, min, max)),
+                       : 0),
```
Joseph-Edwards, line 342: When `number_of_paths` was not a member of the `paths` namespace, this worked fine, but shouldn't have done. This is because both `path-iterators` and `paths` depend on each other. Since this `_count` variable only exists for hacky reasons, I didn't go to the effort of separating out the helpers into different files.

There were other instances where functions in `path-iterators` called functions from `paths`, but the `paths` functions were only convenient wrappers around the `path-iterator` functions anyway, so this was easy to fix. 

It seems like the options we have are:
1. embrace the hack, and factor out `number_of_paths` into its own file that both `paths` and `path-iterators` can depend upon;
2. implement the change described in the hack's FIXME (i.e. properly implement pislo and pstislo iterators); or
3. disregard this PR entirely, and pretend it's not a problem.

@james-d-mitchell, how do you think we should proceed here?

PR #710, opened 2025-03-28
Fix issue in `SimsRefinerIdeals`

james-d-mitchell:
The issue was that we assumed that thread ids were always in the range 0 to max. hardware concurrency + 1, which was probably a bad assumption. The changes in this commit remove this assumption, at the cost of locking a mutex to access the KnuthBendix objects in the refiner. A better fix would be to make `reduce_no_run_no_checks` in `KnuthBendix` thread safe.


james-d-mitchell:
This is a draft because I left some TODOs in the code, just want to see if the CI will run

include/libsemigroups/sims.hpp
```diff
@@ -3120,16 +3120,16 @@ namespace libsemigroups {
    private:
     using node_type    = uint32_t;
     using KnuthBendix_ = KnuthBendix<word_type>;
-    std::vector<KnuthBendix_> _knuth_bendices;
-    Presentation<word_type>   _presentation;
+
+    std::thread::id                                   _default_thread_id;
```
reiniscirpons, line 3124: Do I understand correctly that `_default_thread_id` is just some not necessarily actually used thread id  value used so that we had some standard place to look for an already initialized Knuth-Bendix object? And this shouldn't cause issues if e.g. due to some coincidence `_default_thread_id` is in use, right?

src/sims.cpp
```diff
@@ -1472,54 +1469,72 @@ namespace libsemigroups {
...
+    // Don't copy _knuth_bendices because the thread id's will be wrong
+    _knuth_bendices.emplace(
+        _default_thread_id,
+        (*that._knuth_bendices.find(that._default_thread_id)).second);
+    return *this;
+  }
+
+  SimsRefinerIdeals& SimsRefinerIdeals::operator=(SimsRefinerIdeals&& that) {
+    _default_thread_id = std::move(that._default_thread_id);
+    _knuth_bendices    = std::move(that._knuth_bendices);
```
reiniscirpons, line 1498: In the copy assignment operator we explicitly don't copy the `_knuth_bendices` object since the thread ids will be wrong. Shouldn't we do the same here when move assigning? Since the thread ids will also presumably be wrong?

reiniscirpons (approved):
Works on my system. Haven't checked on any cpu limited platform yet, but if I understand correctly you and Joe had success in reproducing the error and verifying the fix in the docker container?

I had a couple of minor questions about the code but otherwise looks good to me, thanks for the fix James!

include/libsemigroups/sims.hpp
james-d-mitchell: Yes that's correct, according to:

https://en.cppreference.com/w/cpp/thread/thread/id/id

> Default-constructs a new thread identifier. The identifier does not represent a thread.

src/sims.cpp
james-d-mitchell: I don't think we can actually call this operator when there are running threads, so I think this should be ok. The idea being that the lhs of `SimsRefinerIdeals x = std::move(y);` takes over the responsibility for the resources in `y`, so I think this is ok.

PR #715, opened 2025-03-31
Add more `Sims` benchmarks and improve benchmarking infrastructure

reiniscirpons:
This PR adds more benchmarks for the Sims low index congruences algorithm. I plan to add more benchmarks for groups and also to improve the benchmark storage, comparison and conversion as discussed with @james-d-mitchell . The plan is as follows:

- [ ] Add more 1 and 2-sided Sims benchmarks for groups
- [ ] Improve the `to_latex_table.py` script to output either a `rich` table or a `LaTeX` table of benchmarks.
- [ ] Figuring out some infrastructure for comparing benchmarks between different runs.
   - [ ] Find some existing tooling for doing comparisons between benchmark xml files or just benchmark files in general. 
   - [ ]  This may involve adding a `.xsd` xml schema for validation and to handle situations where we change the schema between versions
   - [ ] To change the schema between versions we might need to use some `.xslt` transformations.
 
For the final bullet point the solution discussed was to essentially commit the relevant benchmark run xml files after changes to the appropriate functions and then use some git magic + a benchmark comparison tool to test for performance regressions after changes.

The current PR might only implement a subset of these requirements to facilitate making some benchmark tables for and upcoming paper, but I will open an issue with all the unimplemented bits once its done.


Joseph-Edwards:
The [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark) might be a good point of reference for tooling for doing comparisons between benchmarks. That said, it doesn't solve the issue if we want this to be run locally.

reiniscirpons:
> The [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark) might be a good point of reference for tooling for doing comparisons between benchmarks. That said, it doesn't solve the issue if we want this to be run locally.

This looks like precisely what we want and they also support Catch2. Thanks Joe! To run it locally we can probably just adapt whatever script they are using, I will have a look.

PR #726, opened 2025-04-29
Fix greedy presentation issue

james-d-mitchell:


include/libsemigroups/ukkonen.hpp
```diff
@@ -1063,7 +1064,12 @@ namespace libsemigroups {
       // TODO(later) it'd be better to just convert the values pointed at by the
       // iterators, than to allocate a word_type here, but this is currently a
       // bit tricky to set up
-      add_word(u, word_type(first, last));
+      using points_at_type = std::decay_t<decltype(*std::declval<Iterator>())>;
+      word_type w;
+      for (auto it = first; it != last; ++it) {
+        w.push_back(static_cast<std::make_unsigned_t<points_at_type>>(*it));
+      }
+      add_word(u, w);
```
Joseph-Edwards, line 1072: I think this would be clearer if we threw an exception if any of the points are signed rather than silently converting them to unsigned, unless this specific behaviour is important. In which case, we should add a sentence to the doc explaining this will happen.

Joseph-Edwards (commented):
One small comment, but otherwise this looks good. Thanks @james-d-mitchell!

include/libsemigroups/ukkonen.hpp
Joseph-Edwards: After speaking with @james-d-mitchell, we've decided that doing this casting is necessary. We should, however, check that Ukkonen correctly undoes this operation when returning any points.

james-d-mitchell: I've now updated the code everywhere that accepts iterators in `Ukkonen` to ensure that this casting is done appropriately. Words aren't output by this so there's nothing to update on that front. I still need to update the doc to reflect this change.

james-d-mitchell:
I still need to update the Ukkonen documentation for this one.

james-d-mitchell:
> I still need to update the Ukkonen documentation for this one.

This is now complete.

PR #728, opened 2025-04-30
Make `check_doxygen_line_breaks.py` compatible with Python 3.9

Joseph-Edwards:
This PR makes the `etc/check_doxygen_line_breaks.py` script compatible with Python 3.9 and above.


james-d-mitchell:
Works for me, thanks @Joseph-Edwards

PR #731, opened 2025-05-15
to: fix `to<Presentation>(kb)`

Joseph-Edwards:
This PR refactors/fixes the overload of the `to` function that converts `KnuthBendix` objects to `Presentation` objects. In particular, it properly exposes the overload where we explicitly specify the type of presentation (e.g. `to<Presentation<std::string>>(kb)`), and no longer uses `KnuthBendixImpl`. It also adds some tests.

### DIFF
--- a/include/libsemigroups/to-presentation.tpp
+++ b/include/libsemigroups/to-presentation.tpp
@@ -53,25 +53,29 @@ namespace libsemigroups {
   // KnuthBendix -> Presentation
   ////////////////////////////////////////////////////////////////////////
 
-  template <typename Result, typename Rewriter, typename ReductionOrder>
-  auto to(detail::KnuthBendixImpl<Rewriter, ReductionOrder>& kb)
+  template <typename Result,
+            typename WordIn,
+            typename Rewriter,
+            typename ReductionOrder>
+  auto to(KnuthBendix<WordIn, Rewriter, ReductionOrder>& kb)
       -> std::enable_if_t<
           std::is_same_v<Presentation<typename Result::word_type>, Result>,
           Result> {
-    using Word = typename Result::word_type;
-    if constexpr (std::is_same_v<Word, std::string>) {
-      auto const&               p_orig = kb.internal_presentation();
-      Presentation<std::string> p;
-      p.alphabet(p_orig.alphabet())
-          .contains_empty_word(p_orig.contains_empty_word());
+    using WordOut                      = typename Result::word_type;
+    Presentation<WordIn> const& p_orig = kb.presentation();
+    Presentation<WordIn>        p;
+    p.alphabet(p_orig.alphabet())
+        .contains_empty_word(p_orig.contains_empty_word());
 
-      for (auto const& rule : kb.active_rules()) {
-        presentation::add_rule(p, rule.first, rule.second);
-      }
+    for (auto const& rule : kb.active_rules()) {
+      presentation::add_rule(p, rule.first, rule.second);
+    }
+
+    if constexpr (std::is_same_v<WordIn, WordOut>) {
       return p;
     } else {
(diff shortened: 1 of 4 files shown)


include/libsemigroups/to-presentation.tpp
```diff
@@ -53,25 +53,29 @@ namespace libsemigroups {
...
+    for (auto const& rule : kb.active_rules()) {
+      presentation::add_rule(p, rule.first, rule.second);
+    }
+
+    if constexpr (std::is_same_v<WordIn, WordOut>) {
       return p;
     } else {
       // TODO(1) avoid double copy here
-      return to<Presentation<Word>>(to<Presentation<std::string>>(kb));
+      return to<Presentation<WordOut>>(p);
```
Joseph-Edwards, line 78: I wasn't entirely sure what was being double copied, so I'm not sure if this TODO still applies.

james-d-mitchell: Was just referring to the fact that p is constructed and then immediately thrown away, which still happens so please leave the todo

tests/test-to-presentation.cpp
```diff
@@ -382,4 +382,55 @@ namespace libsemigroups {
...
+
+    if constexpr (std::is_same_v<W1, std::string>) {
+      p.alphabet("hijkl");
+
+      presentation::add_rule(p, "hi", "j");
+      presentation::add_rule(p, "ij", "k");
+      presentation::add_rule(p, "jk", "l");
+      presentation::add_rule(p, "kl", "h");
+      presentation::add_rule(p, "lh", "i");
+    } else if constexpr (std::is_same_v<W1, word_type>) {
```
james-d-mitchell, line 409: Pretty sure this constexpr isn't necessary.

tests/test-to-presentation.cpp
```diff
@@ -382,4 +382,55 @@ namespace libsemigroups {
...
+    kb.run();
+
+    auto q = to<Presentation<W2>>(kb);
+    REQUIRE(q.alphabet().size() == 5);
+    REQUIRE(q.rules.size() == 48);
+
+    if constexpr (std::is_same_v<W1, W2>) {
+      REQUIRE(q.alphabet() == p.alphabet());
+      REQUIRE(q == to<Presentation>(kb));
+    } else if constexpr (std::is_same_v<W2, std::string>) {
```
james-d-mitchell, line 429: Same here

james-d-mitchell (commented):
I'm happy with this, only one quibble, is the test really 4 tests pretending to be one?

Joseph-Edwards:
> I'm happy with this, only one quibble, is the test really 4 tests pretending to be one?

I suppose so. The mathematical presentations are all the same, so I lumped it into one, but it's probably better split. I'll do that now.

tests/test-to-presentation.cpp
```diff
@@ -382,55 +382,144 @@ namespace libsemigroups {
...
+    presentation::sort_each_rule(q);
+    presentation::sort_rules(q);
+    REQUIRE(q.rules
+            == std::vector<std::string>{
+                "hi",  "j",  "hl",  "i",  "ih",  "j",  "ij",  "k",  "ji", "k",
+                "jk",  "l",  "kj",  "l",  "kl",  "h",  "lh",  "i",  "lk", "h",
+                "hhh", "l",  "iii", "h",  "ik",  "hh", "jh",  "hj", "jj", "hk",
+                "jl",  "ii", "kh",  "hk", "ki",  "hh", "kk",  "il", "li", "il",
+                "lj",  "ii", "ll",  "hj", "hhj", "il", "iil", "hhk"});
+  }
```
james-d-mitchell, line 417:
```suggestion
  }
  
```

PR #733, opened 2025-05-16
build: properly set `FMT_HEADER_ONLY`

james-d-mitchell:
This should allow the python bindings to work without setting the flag `--with-external-fmt`.


Joseph-Edwards (commented):
All looks good to me @james-d-mitchell. Everything works on my machine, assuming I have merged these changes into main. In this PR, should we also update the downstream job to drop the `--with-external-fmt` flag, or should we stay as we are as this is what the conda recipe will do?

james-d-mitchell:
> All looks good to me @james-d-mitchell. Everything works on my machine, assuming I have merged these changes into main. In this PR, should we also update the downstream job to drop the `--with-external-fmt` flag, or should we stay as we are as this is what the conda recipe will do?

No I don’t think so, better to use an external format if it’s available, this just makes it possible to use libsemigroups when the included fit is used?

PR #750, opened 2025-07-03
Template `ToWord`

james-d-mitchell:
This is PR number 3 split off from #734 where we make `ToWord` a template.


james-d-mitchell:
Does this look reasonable @Joseph-Edwards ?

PR #755, opened 2025-07-18
Reorganise the doc

Joseph-Edwards:
This PR updates the doc so that it is more similar to https://github.com/libsemigroups/libsemigroups_pybind11. This includes restructuring the sidebar, and making the project info at the top of the sidebar clickable. It also adds some doc that was missing.

As it stands, the **top level** sidebar headers, and some lower level headers, are the same between `libsemigroups` and `libsemigroups_pybind11` (with a small number of exceptions outlined below). I have not checked that the headers at all depths are the same between the two projects. Due to the differences between Sphinx and Doxygen, and Python and C++, I think that it is correct to not insist the sidebar structure be identical. 

There are some notable differences between the doc here and in `libsemigroups_pybind11`:
- `libsemigroups_pybind11` has an "Exceptions" tab in the package info, whereas `libsemigroups` does not, since the exceptions in `libsemigroups` do not offer any insight into the structure of the library (as they do in `libsemigroups_pybind11`). Instead, the `libsemigroups` "Exceptions" tab is in "Data Structures".
- `libsemigroups_pybind11` has an "Enums" tab, whereas `libseimgroups` has a "Types" tab, since we document some none-enum types.
- There are some things not implemented in `libsemigroups_pybind11`, such as `Ranges`. Therefore, there are more tabs in the sidebar of `libsemigroups` than there are in `libsemigroups_pybind11`.

I'm not sure where I stand on the documentation of Enums and types. In `libsemigroups_pybind11`, we have all of the top-level enums documented in the "Enums" tab. I have replicated this behaviour in `libsemigroups`, but something felt wrong about moving the `Order` enum out of the `Orders` tab. I am open to suggestions about which idea is better, because I can see the benefits of both.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,58 +151,6 @@ tracker](https://github.com/libsemigroups/libsemigroups/issues).
 
 We'd like to thank the authors and contributors to these excellent projects!
 
-## Authors
-
-- [Reinis Cirpons][] (<rc234@st-andrews.ac.uk>)
-- [Joseph Edwards][] (<jde1@st-andrews.ac.uk>)
-- [James Mitchell](https://jdbm.me) (<jdm3@st-andrews.ac.uk>)
-
-## Contributors
-
-- [Luna Elliott](https://le27.github.io/L-Elliott/)
-  (<luna.elliott142857@gmail.com>)
-  contributed to the Schreier-Sims implementation.
-- Jan Engelhardt (<jengelh@inai.de>) contributed some bug fixes to the
-  build system, and a number of helpful issues.
-- Ilya Finkelshteyn (<ilyaf@appveyor.com>) contributed to the
-  continuous integration in AppVeyor.
-- Isuru Fernando (<isuruf@gmail.com>) contributed to the build system.
-- [Florent Hivert](https://www.lri.fr/~hivert/)
-  (<Florent.Hivert@lri.fr>) contributed many helpful ideas to
-  `libsemigroups`, and `HPCombi`.
-- [Max Horn](https://math.rptu.de/en/wgs/agag/people/head/prof-dr-max-horn)
-  (<max@quendi.de>) contributed some fixes.
-- [Jerry James](http://www.jamezone.org/) (<loganjerry@gmail.com>)
-  contributed some bugfixes.
-- [Julius Jonušas][] contributed to the implementation of the Froidure-Pin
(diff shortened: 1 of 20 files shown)


james-d-mitchell:
> * `libsemigroups_pybind11` has an "Exceptions" tab in the package info, whereas `libsemigroups` does not, since the exceptions in `libsemigroups` do not offer any insight into the structure of the library (as they do in `libsemigroups_pybind11`). Instead, the `libsemigroups` "Exceptions" tab is in "Data Structures".

Seems sensible, and clear.


> * `libsemigroups_pybind11` has an "Enums" tab, whereas `libseimgroups` has a "Types" tab, since we document some none-enum types.

I'm going to push a change that adds changes the title of the "tab" to be "Enums + Types", mostly because just now I was struggling to find the enum docs (without reading this comment properly first of course).

> * There are some things not implemented in `libsemigroups_pybind11`, such as `Ranges`. Therefore, there are more tabs in the sidebar of `libsemigroups` than there are in `libsemigroups_pybind11`.

There probably should be a `Ranges` tab in the `libsemigroups_pybind11` docs, since if memory serves, we do actually return them in some places.
 
> I'm not sure where I stand on the documentation of Enums and types. In `libsemigroups_pybind11`, we have all of the top-level enums documented in the "Enums" tab. I have replicated this behaviour in `libsemigroups`, but something felt wrong about moving the `Order` enum out of the `Orders` tab. I am open to suggestions about which idea is better, because I can see the benefits of both. 

I think this is fine, although I understand your concern, I'm about to push a change that adds a "see also" to the front "orders" page to make the connection explicit. 

One more thing that I think will be good to replace, I'll comment on in a moment

Joseph-Edwards:
> The other thing: is it possible to remove the "libsemigroups" immediately above "LIBRARY INFO", it seems superfluous given that the logo etc at the top left is now clickable. I tried removing the line:
> 
> 
> 
> ~~~
> 
>     <tab type="mainpage" visible="yes" title="" />
> 
> ~~~
> 
> 
> 
> from `DoxygenLayout.xml` but that appeared to do nothing. This isn't at all important, but if it can be easily removed lets do it.

I had exactly this thought, and tried precisely what you did. I couldn't see an obvious way of getting rid of it unfortunately. I think editing `docs/fixup.py` is probably the way to go.

PR #759, opened 2025-07-20
Bipartition benchmarks

james-d-mitchell:


tests/test-bipart.cpp
```diff
@@ -587,4 +585,58 @@ namespace libsemigroups {
...
+
+    std::unordered_set<Bipartition, Hash<Bipartition>> map;
+
+    for (size_t i = 0; i < 3417; ++i) {
+      map.emplace(bipartition::random(3));
+    }
+    REQUIRE(map.size() >= 200);
+
+    REQUIRE_NOTHROW(bipartition::random(0));
+    REQUIRE(bipartition::random(0) == Bipartition());
```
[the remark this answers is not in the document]
Joseph-Edwards: Removing the occurrences of `bipartition::random(0)` means the tests pass. Same for the benchmarks

tests/test-bipart.cpp
```diff
@@ -587,4 +585,58 @@ namespace libsemigroups {
...
+    auto x = bipartition::random(10);
+    REQUIRE(x.degree() == 10);
+    REQUIRE_NOTHROW(bipartition::throw_if_invalid(x));
+
+    std::unordered_set<Bipartition, Hash<Bipartition>> map;
+
+    for (size_t i = 0; i < 3417; ++i) {
+      map.emplace(bipartition::random(3));
+    }
+    REQUIRE(map.size() >= 200);
```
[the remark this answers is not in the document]
james-d-mitchell: Ah, so 3417 is the number of trials required to get within 99% probability of getting every single such bipartition, and there are 203 of them, so omitting a few for safety's sake.

Joseph-Edwards:
I generated a bunch of data for bipartitions of degree 3, 4 and 5 to have a look at the distribution. For these values of degree, it certainly seems like some bipartitions are preferred more than others.

This plot shows the result of calling `bipartition::random(3)` 101500 times. Each bar represents one of the possible 203 bipartitions. The blue bars record the number of times that bipartition was generated. The orange bars indicate how many times you would expect that bipartition to be generated:

image

It would be interesting to know whether you only get uniformity for large values of degree, or whether we have made a mistake in the implementation

Joseph-Edwards:
This is the same data, but sorted by number of occurrences:
image

Unfortunately, I can't really suggest any reason why this might be. I just thought it was interesting

james-d-mitchell:
Ok, I think I had an off by 1 error, can you check again when you have a minute?

Joseph-Edwards:
This is before your changes:
image

Joseph-Edwards:
I also wrote some tests that perform Pearson's Chi-squared test. Shall I push those @james-d-mitchell? They were not passing before, but they pass with your new changes, so it seems like they are checking for the right thing.

src/bipart.cpp
```diff
@@ -346,6 +372,86 @@ namespace libsemigroups {
...
+            if (std::isnan(num) || std::isnan(den) || std::isnan(num / den)) {
+              LIBSEMIGROUPS_EXCEPTION("the degree ({}) of the argument <x> (a "
+                                      "bipartition) is too large, "
+                                      "please use random(x) instead",
+                                      N / 2);
+            }
+            return num / den;
+          });
+
+      auto                                    M = g(mt) - 1;
```
Joseph-Edwards, line 404: I think this is sometimes causing an underflow error. If I generate enough random bipartitions of degree 1, eventually `g(mt)` will return `0`, so `M` underflows

src/bipart.cpp
```diff
@@ -346,6 +372,86 @@ namespace libsemigroups {
...
+            long double num = std::pow(m, N);
+            long double den = c * factorial(m);
+            if (std::isnan(num) || std::isnan(den) || std::isnan(num / den)) {
+              LIBSEMIGROUPS_EXCEPTION("the degree ({}) of the argument <x> (a "
+                                      "bipartition) is too large, "
+                                      "please use random(x) instead",
+                                      N / 2);
+            }
+            return num / den;
+          });
```
Joseph-Edwards, line 402: I've been looking at the sage implementation of this, and I think this constructor for `std::discrete_distribution` behaves slightly differently to sage's `GeneralDiscreteDistribution`.

From the C++ doc:
> Each of the weights is equal to $w_i$ = unary-op($x_\{min\}$ + δ(i + 0.5))  where $δ = \frac{(x_{max} − x_{min})}{count}$ and $i ∈ \\{0, ..., count − 1\\}$.

In particular, the weight of 0 will not be equal to 0 as it is in sage. I think it is this that is causing the issue with segfault outlined below

src/bipart.cpp
```diff
@@ -36,12 +37,37 @@
...
   // Helpers
   ////////////////////////////////////////////////////////////////////////
 
   namespace {
+    long double factorial(size_t n) {
+      if (n == 0) {
+        return 1;
+      }
+      long double result = 1;
+      for (uint64_t i = 1; i != n; ++i) {
```
Joseph-Edwards, line 50:
```suggestion
      for (uint64_t i = 1; i != n+1; ++i) {
```

Joseph-Edwards:
Due to the random nature of the tests, it seems like sometimes they'll fail. How should we proceed on this? We could either remove the tests completely, or make them as no ci?

james-d-mitchell:
I think maybe we should mark them as not being run in the ci, although it'd be better if we could run them reliably, but this may not be possible.

james-d-mitchell:
I'm going to rebase then I'll merge thanks @Joseph-Edwards

Issue #758, opened 2025-07-20
Changing from `std::string` to `std::basic_string<unsigned char>` in version 3.1.0 is broken


james-d-mitchell:
See: 

https://github.com/conda-forge/libsemigroups-feedstock/pull/91

Exactly the issue that we've discussed before releasing (not sure why it never came up for us, and comes up in the feedstock only, but oh well). This is that `std::char_traits` isn't defined for `unsigned char`, and so we shouldn't have expected this to work. Here are some proposed solutions:

1. Switch back to using `std::string`, and imposing a limit of 128 generators for the time being. **Pros**: easy to do; was the default before; same performance; same memory consumption. **Cons**: limits `KnuthBendix` to 128 generators.
2. Switch to using `std::u16string` (introduced in C++11). **Pros**: easy to do; raises limit on number of generators to `65536`; same performance (on my Mac arm at least). **Cons**: doubles memory consumption.
3. Switch to C++20 which offers `std::u8string`. **Pros**: lets us do exactly as we tried in v3.1.0. **Cons**: is a breaking change and would require a release of v4 (seems a bit drastic)
4. Implement our own `unsigned char` class (i.e. a struct/class with a single data member, and appropriate members fns) + implement `std::char_traits` for this type. **Pros**: allows us to do exactly as we tried in v3.1.0. **Cons**: is complicated. See also https://reviews.llvm.org/D138307#3946939 for some relevant discussion.

Joseph-Edwards:
It's a shame this wasn't caught earlier, and I think it would probably be sensible to try and add a CI job that would have caught this. I'll look into that.

I think 3. is more drastic than the KnuthBendix generator limit necessitates, and 4. would require too much effort for it to be worthwhile. My preference would be something related to 1. and/or 2..

Can we do something similar to what we do with matrices, where the underlying type can be changed depending on the data that needs to be stored?

Alternatively, can we add a template parameter to`KnuthBendixImpl` that dictates what the underlying string type is? Or have a`BigKnuthBendixImpl` that used uint16, and the default class uses std::string?

Another option would be stick with 1., but then add a compiler flag called `--use-prerelease` or `--enable-c++20` or something similar, that would let the user specify that they want to use a later c++ standard, and access KnuthBendix with more than 128 generators, without breaking backwards compatibility (I think).

Joseph-Edwards:
Resolved in #763 and #772.

PR #760, opened 2025-07-21
CI: Add a mac arm runner and update compiler versions

Joseph-Edwards:
This PR adds a job that runs the quick test on the `macos-latest-large` image. If I understand correctly, this currently runs macOS 14 on an arm64 processor. Hopefully, this will fail for the reasons highlighted in #758, and will make it easier to spot problems like this faster moving forwards.

Additionally, this PR unpins the compiler versions that we use to check for compiler warnings.

### DIFF
--- a/.github/workflows/compile-warnings.yml
+++ b/.github/workflows/compile-warnings.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        compiler: [clang++-18, g++-14]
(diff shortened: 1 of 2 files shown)


Joseph-Edwards:
It would seem that `macos-latest-large` runners are only available on GitHub Enterprise or GitHub Team (which we don't have). Never mind.

Joseph-Edwards:
Going to close this for now to do some more testing.

PR #762, opened 2025-07-22
Document + enforce 128 character limit in alphabets for KnuthBendix

james-d-mitchell:
This PR also fixes the annoying compile warnings coming from `forest.hpp`


Joseph-Edwards (commented):
This all looks good to me. We could also use this PR to revert the changes we made for the cygwin specific typedef of `u8string`, since there is now no need to differentiate. If we don't want to make that change, then I am happy to merge this once tests pass.

PR #763, opened 2025-07-22
Knuth bendix 128

james-d-mitchell:


Joseph-Edwards (approved):
All looks good to me. Thanks @james-d-mitchell!

PR #764, opened 2025-07-22
Update compiler versions and add Mac arm runner

Joseph-Edwards:
This PR supersedes #760. It adds a Mac arm runner, and updates the version of the compilers used to check for compiler warnings.

### DIFF
--- a/.github/workflows/compilers.yml
+++ b/.github/workflows/compilers.yml
@@ -14,6 +14,8 @@ jobs:
       fail-fast: false
       matrix:
         sys:
+          # If you update the following list, please also update the compilers
+          # in compiler-warnings.yml to point to the latest version
           - { compiler: g++, version: 9 }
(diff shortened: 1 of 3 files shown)


james-d-mitchell:
> I bumped the version of `clang++` from `clang++-18` to `clang++-19` in the compiler warnings job, but I now realise that doing this requires some extra steps to install the right compiler. This makes me think that it probably isn't best to pin the version at all, and instead just stick with the default.

Agreed, let's do that.

.github/workflows/compilers.yml
```diff
@@ -14,6 +14,8 @@ jobs:
       fail-fast: false
       matrix:
         sys:
+          # If you update the following list, please also update the compilers
```
james-d-mitchell, line 17: This comment doesn't apply anymore right? Please remove.

PR #769, opened 2025-07-23
Fix tests for single core machines

### LINKED ISSUE #767: tests/test-sims.cpp:4587 fails in 1-core-VM
While working on [reproducible builds](https://reproducible-builds.org/) for [openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds), I found that
our libsemigroups-3.1.0 package fails in tests/test-sims.cpp:4587 when run in a 1-core-VM.

Similar to previous issue #251

Maybe just skip the test, if it requires >1 thread.

Joseph-Edwards:
Closes #767.

In a small number of places in the doc, we check the `human_readable_repr` of Sims-related objects. As part of this, we report the number of threads being used.

This PR fixes potential plurality issues (thread vs threads) by always reporting "thread(s)".

### DIFF
--- a/src/sims.cpp
+++ b/src/sims.cpp
@@ -1679,9 +1679,7 @@ namespace libsemigroups {
         needs_and = true;
       }
       result += needs_and ? " and" : "";
-      result += fmt::format(" {} thread{}",
-                            x.number_of_threads(),
-                            x.number_of_threads() == 1 ? "" : "s");
+      result += fmt::format(" {} thread(s)", x.number_of_threads());
       return result;
     }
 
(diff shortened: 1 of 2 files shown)


Joseph-Edwards:
I have checked this locally on an [openSUSE](https://www.opensuse.org/) single-core virtual machine, and the tests now pass. Unfortunately, I couldn't find an easy way of setting up a GitHub runner to only use one core, but I suspect this is probably possible.

Joseph-Edwards:
> I've opened a PR that includes your single-core trick at #770. Hopefully this catches the bug, and then we can check this PR resolves it.

It turns out that it's actually quite hard (or at least non-obvious) to get a GitHub runner to only use one core, so I've closed #770 for now.

james-d-mitchell:
> @bmwiedemann  can you please confirm that this fixes things for you?

Hopefully your other comments have been addressed, we'd be most grateful if you could let us know if the changes in this pr fix the issue you reported @bmwiedemann ?

bmwiedemann:
yes, I had tested it and forgot to comment that it fixed the test on 1-core-VM.

PR #772, opened 2025-07-29
Remove remaining occurrences of `std::basic_string<uint8_t>`

Joseph-Edwards:
This PR builds on #763 in addressing #758. It removes all remaining occurrences of `std::basic_string<uint8_t>`, as they result in undefined behaviour.


james-d-mitchell (approved):
Looks good!

PR #779, opened 2025-08-01
knuth-bendix: add dot function

### LINKED ISSUE #696: Add `dot` function for `KnuthBendix` that draws the Gilman graph + labels
At the moment, we have overloads of the `dot` function that accept a reference to either:
* an `AhoCorasick` instance;
* a `Stephen` instance;
* an `Ukkonen` instance; or
* a `WordGraph` instance.

It would be good to have a an overload for `KnuthBendix` that returns the Gilman graph with labels. Since the Gilman graph is a `WordGraph`, I suspect the `dot(WordGraph<Node>)` would be a good reference for this.

james-d-mitchell:
While doing this I realised that it'd probably be best to update `AhoCorasick` to be a facade for `AhoCorasickImpl`, not sure if we want to do that first or not?

### DIFF
--- a/include/libsemigroups/detail/aho-corasick-impl.hpp
+++ b/include/libsemigroups/detail/aho-corasick-impl.hpp
@@ -32,6 +32,7 @@
 
 #include "libsemigroups/constants.hpp"  // for Undefined, operator!=, UNDEFINED, operator==
 #include "libsemigroups/debug.hpp"      // for LIBSEMIGROUPS_ASSERT
+#include "libsemigroups/dot.hpp"        // for Dot
 #include "libsemigroups/exception.hpp"  // for LIBSEMIGROUPS_EXCEPTION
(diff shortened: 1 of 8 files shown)


james-d-mitchell:
> My unfinished attempt is https://github.com/Joseph-Edwards/libsemigroups/tree/facade-test. The rebase didn't seem like it would be a super quick one, so it didn't seem worthwhile. The tests compile, but they don't all pass.
> 
> One of the things this draft attempts to implement is a `CITOW` style object that doesn't rely on the owner having a `presentation` method. Prompted by this, I think @james-d-mitchell suggested we do some refactoring to create an `Alphabet` object that would make translation things like `CITOW` a bit more general. I don't know if this is something we want to do now too

Thanks @Joseph-Edwards now I remember, I'll have a look at your unfinished attempt, and then comment again here.

PR #785, opened 2025-08-23
Balance presentation

james-d-mitchell:
This function adds:

* a check version of `balance`
* a 2-argument version of `balance` and `balance_no_checks` only taking the inverses and using the entire alphabet for the "letters" argument of the 3-arg version
* a 1-argument version that auto-detects the inverses in rules (only works when they can be unambiguously determined and throws an exception when they cannot be)

a couple of other small things that are related to the above.

Completed on the train from somewhere south of York, and Koln.

### DIFF
--- a/include/libsemigroups/presentation.hpp
+++ b/include/libsemigroups/presentation.hpp
@@ -41,11 +41,11 @@
 #include <utility>           // for move, pair
 #include <vector>            // for vector, operator!=
 
-#include "adapters.hpp"   // for Hash, EqualTo
-#include "constants.hpp"  // for Max, UNDEFINED, operator==
-#include "debug.hpp"      // for LIBSEMIGROUPS_ASSERT
-#include "is_specialization_of.hpp"
-#include "order.hpp"       // for ShortLexCompare
+#include "adapters.hpp"              // for Hash, EqualTo
+#include "constants.hpp"             // for Max, UNDEFINED, operator==
+#include "debug.hpp"                 // for LIBSEMIGROUPS_ASSERT
+#include "is_specialization_of.hpp"  // for is_specialization_of
+#include "order.hpp"                 // for ShortLexCompare
 #include "ranges.hpp"      // for seq, operator|, rx, take, chain, is_sorted
 #include "types.hpp"       // for word_type
 #include "ukkonen.hpp"     // for GreedyReduceHelper, Ukkonen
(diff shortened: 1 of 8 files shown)


Joseph-Edwards (changes requested):
Looks mostly good to me, with a few small changes outlined below. Thanks @james-d-mitchell!

include/libsemigroups/presentation.hpp
```diff
@@ -2158,20 +2273,208 @@ namespace libsemigroups {
...
+    inline void balance_no_checks(Presentation<std::string>&         p,
+                                  std::initializer_list<char> const& inverses) {
+      balance_no_checks(p, p.alphabet(), std::string(inverses));
+    }
+
+    //! \brief Balance the length of the left-hand and right-hand sides.
+    //!
+    //! See
+    //! \ref balance(Presentation<Word1>&, Word2 const&, Word2 const&)
+    //! for details about this function.
```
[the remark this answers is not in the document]
james-d-mitchell: Yeah it should

Joseph-Edwards:
@james-d-mitchell I have made the changes we spoke about, but I feel that removing all of the non-`Word` overloads for balance may break backwards compatibility; it seems the `const char*` overloads existed before. What do you think we should do here? I think my vote is to keep the `const char *` overload and the `Word` overloads, and bin the others?

include/libsemigroups/presentation.hpp
```diff
@@ -2097,6 +2162,48 @@ namespace libsemigroups {
                                       Word const&                 letters,
                                       std::initializer_list<Word> words);
 
+    //! \brief Detect inverses.
+    //!
+    //! This function tries to deduce any inverses defined by the rules of
```
james-d-mitchell, line 2167:
```suggestion
    //! This function tries to deduce group theoretic inverses of letters in the alphabet of \p p defined by the rules of
```

include/libsemigroups/presentation.hpp
```diff
@@ -2097,6 +2162,48 @@ namespace libsemigroups {
...
+    //! presentation in the following way: the rules of the presentation
+    //! where one side has length 2 and the other has length 0 are detected.
+    //! For any such rule we remember that the first letter is the inverse of
+    //! the second and vice versa. If there are no such rules, then no changes
+    //! are made. If there are multiple different such rules and we deduce
+    //! conflicting values for the inverse of a letter, then an exception is
+    //! thrown.
+    //!
```
james-d-mitchell, line 2175: Add: 

> The letters detected to have an inverse are added to \p letters 

(is it cleared first, or are they just appended?)

james-d-mitchell: And a similar statement for `\p inverses`

PR #790, opened 2025-09-02
Check `is_obviously_infinite` before running in `Congruence::number_of_classes`

### LINKED ISSUE #789: `Congruence` runs forever but `KnuthBendix` completes in 80µs
The following seems to run forever:
```cpp
Presentation<std::string> p;
p.alphabet("ab");
presentation::add_rule(p, "bab", "ba");
Congruence c(twosided, p);
REQUIRE(c.number_of_classes() == POSITIVE_INFINITY);
```
whereas the `KnuthBendix` only analogue completes in 80µs:
```cpp
Presentation<std::string> p;
p.alphabet("ab");
presentation::add_rule(p, "bab", "ba");
KnuthBendix kb(twosided, p);
REQUIRE(kb.number_of_classes() == POSITIVE_INFINITY);
```

Joseph-Edwards:
This PR adds a check to see if a `Congruence` object is obviously infinite before calling `run` when computing `Congruence::number_of_classes`.

Closes #789.

### DIFF
--- a/tests/test-cong.cpp
+++ b/tests/test-cong.cpp
@@ -538,6 +538,9 @@ namespace libsemigroups {
     // non-trivial classes.
     // REQUIRE(cong.number_of_non_trivial_classes() == 0);
     REQUIRE(cong.number_of_classes() == POSITIVE_INFINITY);
+    // cong hasn't started yet, because the presentation is obviously infinite.
+    REQUIRE(!cong.started());
+    cong.run();
     REQUIRE(cong.has<KnuthBendix<std::string>>());
     REQUIRE(
         knuth_bendix::non_trivial_classes(*cong.get<KnuthBendix<std::string>>(),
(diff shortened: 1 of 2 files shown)


tests/test-cong.cpp
```diff
@@ -538,6 +538,9 @@ namespace libsemigroups {
     // non-trivial classes.
     // REQUIRE(cong.number_of_non_trivial_classes() == 0);
     REQUIRE(cong.number_of_classes() == POSITIVE_INFINITY);
+    // cong hasn't started yet, because the presentation is obviously infinite.
+    REQUIRE(!cong.started());
+    cong.run();
```
Joseph-Edwards, line 543: Prior to this PR, calling `Congruence::number_of_classes` would always invoke `Congruence::run`. However, with the changes in this PR, there are instances where this does not happen (namely when the given presentation is obviously infinite). As such, some of the tests that required `Congruence::run` to have been called began to fail.

I addressed this by directly calling `Congruence::run` but otherwise left everything else unchanged. Perhaps it would be better to restructure the tests. I'm open to any opinions on this one.

james-d-mitchell:
This probably should be rebased after the merge of #791 so that the ci can pass.

PR #797, opened 2025-09-16
Use the latest version of backward

### LINKED ISSUE #796: Compilation failure on FreeBSD: no member named 'begin' in 'backward::StackTrace'
Originally reported by @yurivict in https://github.com/gap-system/gap/issues/6100

```
In file included from src/exception.cpp:33:
/usr/ports/math/gap/work/gap-4.15.0-beta1/pkg/semigroups/libsemigroups/extern/backward-cpp/backward.hpp:1337:25: error: no member named 'begin' in 'backward::StackTrace'
1337 |       load_addresses(st.begin(), static_cast<int>(st.size()));
|                      ~~ ^
src/exception.cpp:50:8: note: in instantiation of function template specialization 'backward::TraceResolverImplBase::load_stacktrace<backward::StackTrace>' requested here
50 |     tr.load_stacktrace(st);
|        ^
1 error generated.
```

Later I commented this line out and was able to build it, but the issue is still there.

Version: 4.15.0-beta1
clanng-19
FreeBSD 14.3

Joseph-Edwards:
This PR updates the version of [backward-cpp](https://github.com/bombela/backward-cpp) to be the same as the current head of their master branch.

In particular, these changes include https://github.com/bombela/backward-cpp/commit/bbe537f408f45c42549ede487a1034d39d9740a3, which should fix #796.

### DIFF
--- a/src/exception.cpp
+++ b/src/exception.cpp
@@ -59,22 +59,24 @@ namespace libsemigroups {
     st.load_here();
     backward::TraceResolver tr;
     tr.load_stacktrace(st);
-    for (size_t i = st.size() - 1; i-- > 0;) {
-      backward::ResolvedTrace trace = tr.resolve(st[i]);
-      size_t                  pos   = trace.object_function.find("(");
-      if (pos == std::string::npos) {
-        continue;
+    if (st.size() > 0) {
+      for (size_t i = st.size() - 1; i-- > 0;) {
+        backward::ResolvedTrace trace = tr.resolve(st[i]);
+        size_t                  pos   = trace.object_function.find("(");
+        if (pos == std::string::npos) {
+          continue;
+        }
+        std::string func(trace.object_function.begin(),
+                         trace.object_function.begin() + pos);
+        if (func.find("Catch") != std::string::npos
+            || func.find("C_A_T_C_H") != std::string::npos
+            || func.find("LibsemigroupsException") != std::string::npos
+            || func.find("pybind11") != std::string::npos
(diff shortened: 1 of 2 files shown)


Joseph-Edwards:
@james-d-mitchell proposed that we include the version of `backward` in our file path. However, since the latest release (v1.6) was over four years ago, and several changes have been made since then, it felt slightly misleading to call our version v1.6. I think it would be a good idea to include the version number in the path if they release v1.7.

Joseph-Edwards:
> Probably a separate issue but can we also add a ci jobs for openbsd?

Yeah this sounds sensible. Should be easy enough to do with the use of something like https://github.com/marketplace/actions/freebsd-vm. Famous last words.

I'll open a separate PR that should probably get merged before this one so we can check this PR behaves as expected.

Joseph-Edwards:
It seems as though, either, the tests are taking too long to run, or something is incorrectly hanging. I'll investigate

james-d-mitchell:
It looks like no output from the actual tests, so either they're failing to run, or failing before any output, or it's taking an hour to build, and then getting cancelled. Is it possible that the whole thing is running in a vm and so it's super slow?

james-d-mitchell:
Looks like it was probably the last thing in my list, I've doubled the time out to 2 hours, if it really takes that long then I think it might be better to only run these when we are making a release, not for every PR, because 2 hours is a long time.

james-d-mitchell (approved):

PR #798, opened 2025-09-16
Add an option to compile with external backward

Joseph-Edwards:
This PR adds the compiler flag `--with-external-backward` to enable compiling with an external version of backward.

It's not clear to me how backward really works, so I don't know whether the tests are thorough enough. I'm not familiar with any way to test the output of a compiler error in the CI.

### DIFF
--- a/.github/workflows/config-options.yml
+++ b/.github/workflows/config-options.yml
@@ -56,6 +56,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        external-backward:
+          - --with-external-backward
+          - ""
         external-fmt:
           - --with-external-fmt
           - ""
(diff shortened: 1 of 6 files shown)


james-d-mitchell:
Thanks @Joseph-Edwards I've tried this, and it works just fine. It's not as convenient to use as `--with-external-eigen` because we have to specify the include path manually, whereas `--with-external-eigen` uses `pkg-config`. However, it seems that this isn't something we can fix ourselves, since there's no `.pc` file for `backward` (at least it's not installed by conda/mamba). 

I've opened an issue about that: 

https://github.com/conda-forge/backward-cpp-feedstock/issues/25

james-d-mitchell:
Basically I'm happy to merge this once the CI passes.

PR #799, opened 2025-09-16
Add CI jobs for OpenBSD and FreeBSD

Joseph-Edwards:
This PR adds workflow jobs that run the quick tests on FreeBSD and OpenBSD


Joseph-Edwards:
The new jobs will fail until #797 is merged.

PR #806, opened 2025-09-22
Manually brew install ccache

### LINKED ISSUE #805: The MacOS-x86 CI jobs are timing out
This is currently blocking #735 and #754.

It seems as though these jobs are timing out whilst installing `ccache`. The error messages seem to suggest this might be because we are using macOS-13 for this job (this is how we can run tests on x86 architecture).

I'm doing some experimenting now, but it might be sensible to drop `ccache` for this job and see if that fixes things.

Joseph-Edwards:
For some reason, [setup-ccache-action](https://github.com/Chocobo1/setup-ccache-action) has become quite slow on our macos-13 runner. This PR manually changes that job so that `ccache` is manually installed, and then uses setup-ccache-action to do the configuration.

Closes #805

### DIFF
--- a/.github/workflows/os.yml
+++ b/.github/workflows/os.yml
@@ -75,11 +75,13 @@ jobs:
       CXXFLAGS: -fdiagnostics-color
     steps:
       - uses: actions/checkout@v5
+      - name: Install ccache
+        run: brew install ccache
       - name: Setup ccache . . .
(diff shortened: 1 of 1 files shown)


Joseph-Edwards:
I've opened [this issue](https://github.com/Chocobo1/setup-ccache-action/issues/22) on the action, so hopefully in the future we can revert back to the old, more concise, solution.

PR #815, opened 2025-09-25
Don't rely on the `standalone` package to convert the `to` table to an svg

### LINKED ISSUE #814: Removed inkscape argument in `standalone` pack for to-table svg
The [`standalone`](https://github.com/MartinScharrer/standalone) pack used for generating the to-table svg is currently using the `--pdf-page` argument which was removed in Inkscape version 1.3. There was no backwards compatibility introduced (seen [here](https://gitlab.com/inkscape/inkscape/-/commit/b6a32b7d851cfd2cbb1987a6b9b6392680be6a03#7602770f447739856d214b16baf5e192bc03de99)), so if your inkscape version is more recent the svg conversion won't complete. 

The issue is documented upstream (see [here](https://github.com/MartinScharrer/standalone/issues/19)) and a [pull request](https://github.com/MartinScharrer/standalone/pull/21) has been opened, so it may get fixed for us. However, it doesn't look like the repository is being actively maintained so it might be worth looking into a temporary fix.   

If needing to generate the to-table svg, after running `make doc` go into `docs/pictures` and run:

```shell
inkscape --pdf-poppler --export-type="svg" --pages=1 -o to-table.svg to-table.pdf
```

Joseph-Edwards:
This PR separates the `to` table construction into two steps. The first is to build the `to` table as a pdf, and the second converts that pdf to an svg. This way, we don't rely on the conversion function specified by the `standalone` package that seemingly relies on some deprecated flag.

The flag in question is `--pdf-page`, but this doesn't seem to be needed anyway, so we remove it.

Closes #814.

### DIFF
--- a/etc/make-doc.sh
+++ b/etc/make-doc.sh
@@ -7,6 +7,12 @@ if command -v pdflatex && command -v inkscape 2>&1 >/dev/null; then
     echo "Building to-table . . ."
     pdflatex to-table.tex
     inkscape --pdf-poppler --export-type="svg" -o to-table.svg to-table.pdf
+    if [ $? -eq 0 ]; then
+        echo Successfully created to-table.svg
+    else
+        echo Warning: there was an issue with inkscape when creating to-table.svg
+        echo Please check the state of to-table.svg
+    fi
     cd ../..
 else
     echo "Not building to-table"


Joseph-Edwards:
> Only thing maybe worth adding is a log statement to confirm the svg was generated from the return code of the inkscape command, as it is mostly silent.

Yeah that's a good idea. I'll add that.

> Also if you want to get rid of that one warning message you can add `--pages=1`:

Since the version of inkscape that is installed by default on some Linux distros (such as Ubuntu) doesn't yet support the `--pages` flag, I think having the warning for now is preferable over requiring users to update some software that they might not even realise they have installed.

PR #820, opened 2025-09-29
Use unique identifier for test timing

Joseph-Edwards:
This PR changes the behaviour of the `generate-test-times.py` script so tests are selected based on their `LIBSEMIGROUPS_TEST_PREFIX`. This is the tag we use to warn the user of duplicated test case names, so it is our best bet to uniquely identify a test.

It also adds some timing data for `test_all`.


james-d-mitchell:
> We should fix that test to be better, the one that caused the conda ci to fail just now.

Done in the rc-v3.2.0 branch

james-d-mitchell:
@Joseph-Edwards I took the liberty of dropping it for you :)

PR #835, opened 2025-10-16
forest: add validation

james-d-mitchell:
This commit adds the functions:

- forest::is_forest (helper)
- throw_if_not_acyclic (mem fn)

to the class Forest, so that we can check that a Forest instance is well-defined (no cycles). These functions are now called in appropriate places as assertions in no_checks functions and so that exceptions are thrown in the checks versions.

### DIFF
--- a/src/forest.cpp
+++ b/src/forest.cpp
@@ -18,18 +18,15 @@
 
 #include "libsemigroups/forest.hpp"
 
-#include <algorithm>         // for fill
-#include <cstddef>           // for size_t
-#include <initializer_list>  // for initializer_list
-#include <iosfwd>            // for ostream
-#include <vector>            // for vector, allocator, operator==
+#include <algorithm>  // for fill, count
+#include <cstddef>    // for size_t
+#include <vector>     // for vector
 
-#include "libsemigroups/bipart.hpp"
-#include "libsemigroups/constants.hpp"  // for Max, operator!=, operator==
-#include "libsemigroups/exception.hpp"  // for LIBSEMIGROUPS_EXCEPTION
-#include "libsemigroups/types.hpp"      // for word_type
-
-#include "libsemigroups/detail/string.hpp"  // for operator<<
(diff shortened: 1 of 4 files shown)


src/forest.cpp
```diff
@@ -131,6 +128,74 @@ namespace libsemigroups {
...
+                                  cycle);
+        } else {
+          LIBSEMIGROUPS_EXCEPTION("the Forest object contains the {}"
+                                  " and is invalid",
+                                  cycle);
+        }
+      }
+      current++;
+    }
+  }
```
Joseph-Edwards: If this is performance critical, I think we can improve the complexity from $\mathcal{O}(n^2)$ to $\mathcal{O}(n + e)$ by doing a topological sort here instead. A directed graph is acyclic if and only if it can be topologically sorted.

If detecting the location of the cycles is important too, [this StackOverflow post](https://stackoverflow.com/questions/583876/how-do-i-check-if-a-directed-graph-is-acyclic) tells me that [Tarjan's Algorithm](https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm) is similar to a topological sort and can detect where the cycles are.

Joseph-Edwards (commented):
This looks good to me. I've left a comment that is only relevant if the complexity of `throw_if_not_acyclic` and `is_forest` is important. If not, then I'm happy for this to be merged.

Thanks @james-d-mitchell!

src/forest.cpp
james-d-mitchell: I'm pretty sure the complexity of the implementation is $O(n)$ where $n$ is the number of nodes.

Joseph-Edwards: Yeah. I incorrectly interpreted this as:
```
for each node:
  follow the parent until you get to the root or find a repeat
```
which I think is almost what happens, except we might stop early if we see a node we've already explored.

I don't quite understand why `seen` needs to be a vector of integers rather than a vector of bools, but that isn't really important.

PR #839, opened 2025-10-29
Add weighted shortlex

jswent:


jswent:
> Please also add some documentation for the new functions you've added.

For sure, just wanted to make sure the code is on the right track. I'll get the full PR done up before our meeting if I hear from James by then.

Joseph-Edwards:
> this looks like a problem with the latest arm runner

Yeah I agree. I'll look into it

include/libsemigroups/order.hpp
```diff
@@ -283,7 +286,8 @@ namespace libsemigroups {
   //!
   //! Defined in `order.hpp`.
   //!
-  //! Compare two objects of the same type using \ref shortlex_compare.
+  //! This function compares two objects of the same type using \ref
+  //! shortlex_compare.
```
james-d-mitchell:
```suggestion
  //! This function compares two objects of the same type using 
  //! \ref shortlex_compare.
```

include/libsemigroups/order.hpp
```diff
@@ -550,6 +558,453 @@ namespace libsemigroups {
...
+  //! std::vector<size_t> const&).
+  template <typename T, typename = std::enable_if_t<!rx::is_input_or_sink_v<T>>>
+  bool wtshortlex_compare_no_checks(T const&                   first1,
+                                    T const&                   last1,
+                                    T const&                   first2,
+                                    T const&                   last2,
+                                    std::vector<size_t> const& weights) {
+    size_t weight1 = 0;
+    for (auto it = first1; it != last1; ++it) {
+      weight1 += weights[*it];
```
james-d-mitchell: Could replace this with `std::accumulate(first1, last1, 0)`

jswent:
This should fix the sanitizer errors. I added a test which checks that running a bad input with no checks doesn't throw an exception, but obviously that indexes out of bounds which causes valgrind and asan to error. I could ignore that specific case, but I figured it better to just remove altogether as it's largely superfluous.

PR #844, opened 2025-11-05
Improve valgrind error checking

Joseph-Edwards:
This PR improves the valgrind CI job. Previously, the code was checking the output of valgrind for the string "invalid" or "lost", whereas now we actually set a non-zero exit code if valgrind reports an error.

It also adds the `no-valgrind` tag to some of the newer tests that run for a long time in valgrind mode.

This addresses an issue identified by @jswent in #839.

An example of this correctly failing on a memory leak is available in this [dummy CI job](https://github.com/Joseph-Edwards/libsemigroups/actions/runs/19107493616/job/54595273053)


jswent:
Thanks @Joseph-Edwards

PR #846, opened 2025-11-06
CI: Don’t sudo pip install

Joseph-Edwards:
This PR removes the sudo prefix from the pip commands in the CI, as this is generally discouraged.


james-d-mitchell:
Thanks @Joseph-Edwards

PR #855, opened 2025-11-17
Refactor some transformations code to better support HPCombi in the python bindings

james-d-mitchell:


james-d-mitchell:
I think I've done everything I want in this PR, there are still some issues with the documentation of the adapters for `HPCombi` element types (missing doc for parameters etc), but those existed before so and I got bored adding them, and gave up. So propose to merge as is, pending any comments you might have @Joseph-Edwards and the ci passing of course.

james-d-mitchell:
I think the failing CI jobs can be ignored (they relate to failing to upload coverage reports to codecov)

include/libsemigroups/hpcombi.hpp
```diff
@@ -1072,17 +1294,27 @@ namespace libsemigroups {
                         std::vector<uint8_t>(ran.begin(), ran.end()),
                         M);
   }
+#else
+  template <typename Int>
+  [[nodiscard]] HPCombi::PPerm16
+  make<HPCombi::PPerm16>(std::initializer_list<Int> const& dom,
+                         std::initializer_list<Int> const& ran,
+                         size_t                            deg = 16);
```
Joseph-Edwards: I assume that we add the `#ifndef LIBSEMIGROUPS_PARSED_BY_DOXYGEN` to get around the unsightly `enable_if_is_same_stuff`. If this is the only reason, then the arguments of this doxygen-only declaration should be the same as those in the actual declaration.

It's not clear to me which of `size_t deg = 16` or `size_t M` is correct.

include/libsemigroups/transf.hpp
```diff
@@ -957,23 +953,9 @@ namespace libsemigroups {
...
-      if (val >= M && val != UNDEFINED) {
-        LIBSEMIGROUPS_EXCEPTION("image value out of bounds, expected value in "
-                                "[{}, {}), found {}",
-                                0,
-                                M,
-                                val);
-      }
-    }
-  }
+  [[deprecated]] void
```
Joseph-Edwards: We should add a deprecation warning to the doc like we have done elsewhere

include/libsemigroups/transf.hpp
```diff
@@ -1358,9 +1334,9 @@ namespace libsemigroups {
   //! \complexity
   //! Linear in \c f.degree().
   template <size_t N, typename Scalar>
-  void throw_if_not_pperm(PPerm<N, Scalar> const& f) {
+  [[deprecated]] void throw_if_not_pperm(PPerm<N, Scalar> const& f) {
```
Joseph-Edwards: Same as above

PR #857, opened 2025-11-19
Add max-plus implementation for Konieczny

flsmith:


james-d-mitchell:
Super, thanks @flsmith! There's something seriously wrong with the reporting, if it's enabled in test case 907, then lots of blank lines are reported. This might be because of something we did elsewhere (i.e. we broke the reporting). 

The reasons for the failing tests (except possibly the one that timed out) should be easy to fix.

include/libsemigroups/matrix.hpp
```diff
@@ -7867,6 +7869,70 @@ namespace libsemigroups {
...
+
+      for (size_t i = 0; i < rows.size(); ++i) {
+        auto rv = typename Mat::RowView(rows[i]);
+        rvs.push_back(rv);
+        lookup.insert({&(*rv.begin()), i});
+      }
+      std::decay_t<Container> result;
+      for (auto rv : row_basis<Mat>(rvs)) {
+        // TODO: really???
+        result.push_back(std::move(rows[lookup.at(&(*rv.begin()))]));
```
james-d-mitchell:
```suggestion
        auto&& row = rows[lookup.at(&(*rv.begin()))];
        result.push_back(std::forward<decltype(row)>(row));
```

PR #858, opened 2025-11-19
Add twisted bipartition implementation for Konieczny

flsmith:


james-d-mitchell (changes requested):
Thanks @flsmith, I've added a bunch of TODOs to the code, and made some minor modifications. Perhaps we can discuss this afternoon?

Issue #788, opened 2025-09-02
Add "manual examples" to test file(s)


james-d-mitchell:
I'm imagining something like autodoc for GAP does, where we run a script that takes the code examples out of the documentation and puts them in a test file. This should be more or less straightforward to achieve using doxygen, and some of the existing parsing tools.

PR #859, opened 2025-11-19
Add "manual examples" to test file : 0.1.0

jfjameslee:
See Issue #788. 

This PR consists of a python script that extracts, minimally parses and formats all "runnable" code examples in the docs comments for /include/libsemigroups. 

### TLDR
Run `make doc-test` to extract, compile and test any "runnable" code examples in the libsemigroups documentation. `\skip-test` and `//->` are special patterns explained below.

### Step 0 - Always Exclude

In some instances, e.g. for the current v4 namespace, it is necessary to skip an entire file by default. See `SKIP_ALWAYS` defined in the constants field of the python code block extraction script.

### Step 1 - Code Block Extraction
Running the extraction script:
`> python3 ./etc/extract-doc-examples.py ./include/libsemigroups`
Extraction that recursively explores folders:
`> python3 ./etc/extract-doc-examples.py ./include/libsemigroups -r`
Extraction that excludes specific files:
`> python3 ./etc/extract-doc-examples.py ./include/libsemigroups -e forest.hpp`

**The extraction script will skip over any code blocks marked \skip-test**
``` 
//! \par Example from adapters.hpp line 120
//! \code
//! \skip-test
//! template <>
//! struct Complexity<KBE> {
//!   constexpr size_t operator()(KBE const&) const noexcept {
//!     return LIMIT_MAX;
//!   }
//! };
//! \endcode
```
**The extraction script will add an assertion for function return values where expected output is marked with `//->`:**
```
//! \par Example from schreier-sims.hpp line 166
//! \code
//!  SchreierSims<5> S;
//!  using Perm = decltype(S)::element_type;
//!  S.add_generator(Perm({1, 0, 2, 3, 4}));
//!  S.add_generator(Perm({1, 2, 3, 4, 0}));
//!  S.size(); //-> 120
//!  \endcode
```

### Step 2 - Compile
Run `> make test_docs_code_examples` to compile the set of code examples extracted. This also validates all of the runnable code examples (i.e. they are syntactically valid).

### Step 3 - Run
Execute the test file to verify that all the runnable code blocks are correct.
`> ./test_docs_code_examples`

### Important files
- [Code Block Extraction File](../etc/extract-doc-examples.py)
- [Code Block Test File](../tests/test-docs-code-examples.cpp)

### DIFF
--- a/include/libsemigroups/order.hpp
+++ b/include/libsemigroups/order.hpp
@@ -508,7 +508,7 @@ namespace libsemigroups {
   //! \param y pointer to the second object for comparison.
   //!
   //! \returns The boolean value \c true if the value pointed to by \p x is less
-  //! than the value pointed to by \p y with respect to the recursive path
+  //! than the value pointed to by \p y with r to the recursive path
   //! ordering, and \c false otherwise.
   //!
   //! \exceptions
@@ -655,6 +655,7 @@ namespace libsemigroups {
   //!
   //! \par Possible Implementation
   //! \code
+  //! \skip-test
   //! wt_shortlex_compare_no_checks(
   //!   x.cbegin(), x.cend(), y.cbegin(), y.cend(), weights);
   //! \endcode
@@ -702,6 +703,7 @@ namespace libsemigroups {
   //!
   //! \par Possible Implementation
   //! \code
+  //! \skip-test
   //! wt_shortlex_compare_no_checks(
   //!   x->cbegin(), x->cend(), y->cbegin(), y->cend(), weights);
   //! \endcode
@@ -827,6 +829,7 @@ namespace libsemigroups {
   //!
   //! \par Possible Implementation
   //! \code
+  //! \skip-test
   //! wt_shortlex_compare(
   //!   x.cbegin(), x.cend(), y.cbegin(), y.cend(), weights);
   //! \endcode
@@ -873,6 +876,7 @@ namespace libsemigroups {
   //!
   //! \par Possible Implementation
   //! \code
+  //! \skip-test
   //! wt_shortlex_compare(
   //!   x->cbegin(), x->cend(), y->cbegin(), y->cend(), weights);
   //! \endcode
@@ -1097,6 +1101,7 @@ namespace libsemigroups {
   //!
   //! \par Possible Implementation
   //! \code
+  //! \skip-test
   //! wt_lex_compare_no_checks(
   //!   x.cbegin(), x.cend(), y.cbegin(), y.cend(), weights);
   //! \endcode
@@ -1144,6 +1149,7 @@ namespace libsemigroups {
   //!
   //! \par Possible Implementation
   //! \code
+  //! \skip-test
   //! wt_lex_compare_no_checks(
   //!   x->cbegin(), x->cend(), y->cbegin(), y->cend(), weights);
   //! \endcode
@@ -1269,6 +1275,7 @@ namespace libsemigroups {
   //!
   //! \par Possible Implementation
   //! \code
+  //! \skip-test
   //! wt_lex_compare(
   //!   x.cbegin(), x.cend(), y.cbegin(), y.cend(), weights);
   //! \endcode
@@ -1314,6 +1321,7 @@ namespace libsemigroups {
   //!
   //! \par Possible Implementation
   //! \code
+  //! \skip-test
   //! wt_lex_compare(
   //!   x->cbegin(), x->cend(), y->cbegin(), y->cend(), weights);
   //! \endcode


include/libsemigroups/word-range.hpp
```diff
@@ -2008,6 +2009,7 @@ namespace libsemigroups {
   //! in a compact form.
   //! \par Example
   //! \code
+  //! \skip-test
   //! 012_w      // same as word_type({0, 1, 2})
```
james-d-mitchell: we could test these 
```suggestion
  //! 012_w      //-> word_type({0, 1, 2})
```

etc/extract-doc-examples.py
```diff
@@ -0,0 +1,366 @@
...
+                        function_returns = parts[1].strip()
+
+                        #  validate format of return
+                        #  TODO: This
+
+                        #  Add require value eq
+                        current_block.append(
+                            f"  REQUIRE({function_call} == {
+                                function_returns});"
+                        )
```
Joseph-Edwards: This is only valid if the function call takes up exactly one line, which isn't always the case. For example:
``` c++
//! freeband_equal_to({0, 1, 2, 3, 4, 0, 1, 2, 3, 4},
//!                   {4, 3, 2, 1, 0, 4, 3, 2, 1, 0});    //-> false
```
would get (incorrectly) converted to
```c++
freeband_equal_to({0, 1, 2, 3, 4, 0, 1, 2, 3, 4},
REQUIRE({4, 3, 2, 1, 0, 4, 3, 2, 1, 0}) == false);
```
rather than
```c++
REQUIRE(freeband_equal_to({0, 1, 2, 3, 4, 0, 1, 2, 3, 4},
                          {4, 3, 2, 1, 0, 4, 3, 2, 1, 0})
        == false);
```

Without writing a parser for C++, it's not clear to me how to nicely allow multi-line function calls. Instead, I think we should probably add a brief comment explaining that this is only valid for one-line function calls. Then, in the code blocks, we should try and avoid long lines that have associated return values. For example
``` c++
    //! freeband_equal_to({0, 1, 2, 3, 4, 0, 1, 2, 3, 4},
    //!                   {4, 3, 2, 1, 0, 4, 3, 2, 1, 0});    //-> false
```
could become
``` c++
    //! word_type x = {0, 1, 2, 3, 4, 0, 1, 2, 3, 4};
    //! word_type y = {4, 3, 2, 1, 0, 4, 3, 2, 1, 0};
    //! freeband_equal_to(x, y);    //-> false
```

@james-d-mitchell might have some good suggestions on this one too.

etc/extract-doc-examples.py
```diff
@@ -0,0 +1,366 @@
...
+
+    if total_blocks == 0:
+        print("No code blocks found in any files.")
+    else:
+        print(f"Total code blocks found: {total_blocks}")
+
+
+########################################################################
+# File Formatting
+#########################################################################
```
Joseph-Edwards: Rather than having this formatting section, it might be better to add a recipe at the bottom of `Makefile.in` that generates the doc tests, and then formats the generated files. Something like
```makefile
doc-tests:
	@echo "Extracting code blocks to test. . .";
	python3 etc/extract-doc-examples.py
    clang-format-15 tests/test-docs-code-exampes.cpp
```

This way, you can still run this python file without a formatter, and then format it yourself. Helpful if you have your `clang-format` installed under a different name/alias.

include/libsemigroups/freeband.hpp
```diff
@@ -76,7 +76,7 @@ namespace libsemigroups {
   //! freeband_equal_to({0, 1, 2, 3, 2, 1, 0},
   //!                   {0, 1, 2, 3, 2, 3, 2, 1, 0}); // true
```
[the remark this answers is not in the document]
Joseph-Edwards: We also need to fix the long lines (re: my earlier comment)

Joseph-Edwards (changes requested):
As I've said several times, I think the changes in this PR are awesome, and will prove very helpful down the line.

It seems like there are some formatting issues in some of the files you've edited, which can be fixed by running `make format` (perhaps after making `clang-format-15` an alias for `clang-format@15`).

Really nice work Josh! :)

etc/extract-doc-examples.py
Joseph-Edwards: Ah my bad. It should be added to `Makefile.am`. `Makefile.in` gets generated from `Makefile.am` when `./configure` is run.

include/libsemigroups/matrix.hpp
```diff
@@ -3812,9 +3816,9 @@ namespace libsemigroups {
   //! \par Example
   //! \code
   //! // default construct an uninitialized 3 x 3 static matrix
-  //! BMat<3> m;
+  //! [[maybe_unused]] BMat<3> m3;
```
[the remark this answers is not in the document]
josh-dev-things: Agreed

josh-dev-things:
I believe I have reviewed & addressed all of the above properly? I am happy to add more fixes / improvements where required.

james-d-mitchell:
@josh-dev-things can you please rebase onto `main`?

josh-dev-things:
@james-d-mitchell Should all be fine now I think 👍

Issue #842, opened 2025-11-04
Bug dans le `Paths` objet: Unclear documentation of `max` property behavior.


reiniscirpons:
Consider the following code
```python3
>>> from libsemigroups_pybind11 import WordGraph, Paths, Order, UNDEFINED
>>> wg = WordGraph(2, [[0, 1], [0, UNDEFINED]])
>>> list(Paths(wg).source(0).target(0).max(4))
[[0], [0, 0], [1, 0], [0, 0, 0], [0, 1, 0], [1, 0, 0]]
```
It seems that the `max` property actually gives you all the words of length up to `max-1` instead of `max`. As it stands, the documentation says

> Set the maximum length of path in the range.
>
> This function can be used to set the maximum length of path that will be contained in the range. If this function is not called, then the
> range will contain paths of unbounded length (possibly infinitely many).

which to me reads like `max` should be included. But maybe, like in other parts of the library, the endpoint is not included, but we should clarify if this is the case or fix it if this is a bug. 

Sibling bug to libsemigroups/libsemigroups#841 .

PR #865, opened 2025-11-25
Refactor and fix issues with Path iterators

### LINKED ISSUE #841: Bug dans le `Paths` objet: Missing empty word depending on order
Consider the following code
```python3
>>> from libsemigroups_pybind11 import WordGraph, Paths, Order, UNDEFINED
>>> wg = WordGraph(2, [[0, 1], [0, UNDEFINED]])
>>> [] in list(Paths(wg).source(0).target(0).max(4))
False
>>> [] in list(Paths(wg).source(0).target(0).max(4).order(Order.lex)))
True
```
It seems that depending on the ordering used, the `Paths` object may or may not include the empty word as a valid path. I think for the particular word graph `wg` above, the empty word should be included, so that the list of words returned by the paths object should be:
```python3
[
  [],
  [0],
  [0, 0],
  [0, 0, 0],
  [0, 1, 0],
  [1, 0],
  [1, 0, 0]
]
```

Sibling bug to libsemigroups/libsemigroups#842 .

james-d-mitchell:
This PR resolves #841 #842 by refactoring some of the code in `path-iterators.*pp`, and supersedes #708.

### DIFF
--- a/include/libsemigroups/paths-count.tpp
+++ b/include/libsemigroups/paths-count.tpp
@@ -99,7 +99,7 @@ namespace libsemigroups::v4::paths {
       uint64_t const N     = wg.number_of_nodes();
       auto           acc   = matrix::pow(am, min);
       uint64_t       total = 0;
-      for (size_t i = min; i < max; ++i) {
+      for (size_t i = min; i <= max; ++i) {
         uint64_t add = std::accumulate(acc.cbegin() + source * N,
                                        acc.cbegin() + source * N + N,
                                        static_cast<uint64_t>(0));


include/libsemigroups/num-paths.tpp
```diff
@@ -0,0 +1,468 @@
...
+        }
+        total += add;
+        acc *= am;
+      }
+#else
+      auto           tmp   = am;
+      uint64_t const N     = wg.number_of_nodes();
+      auto           acc   = matrix::pow(am, min);
+      uint64_t       total = 0;
+      for (size_t i = min; i < max; ++i) {
```
Joseph-Edwards:
```suggestion
      for (size_t i = min; i <= max; ++i) {
```

Joseph-Edwards: This ought to fix the failing `--disable-eigen` tests.

Joseph-Edwards (changes requested):
Apart from the small `<` -> `<=` change, and some doc/filename suggestions, this looks good to me (assuming we can get the other tests passing). Thanks @james-d-mitchell.

include/libsemigroups/detail/path-iterators.hpp
```diff
@@ -26,18 +26,20 @@
 #include <vector>       // for vector, vector<>::const...
 
 #include "libsemigroups/constants.hpp"           // for UNDEFINED, Undefined
-#include "libsemigroups/detail/iterator.hpp"     // for default_postfix_incre...
+#include "libsemigroups/paths-count.hpp"         // for number_of_paths
```
Joseph-Edwards:
```suggestion
#include "libsemigroups/paths-count.hpp"         // for paths::count
```

james-d-mitchell:
Am I ok to merge @Joseph-Edwards ?

PR #867, opened 2025-11-27
Add weighted lexicographical compare

jswent:
This PR adds support for weighted lexicographical comparison using `std::lexicographical_compare`, with implementation nearly identical to the weighted shortlex comparison from #839. 

Also based off of the [Word-ordering methods](https://maffsa.sourceforge.net/manpages/word_ordering_methods.html) reference from [MAF](https://maffsa.sourceforge.net/manpages/MAF.html)


james-d-mitchell (approved):
Looks great, thanks @jswent

PR #871, opened 2025-12-04
Update copyright

Joseph-Edwards:
The PR updates the copyright info some of the files.

@josh-dev-things and @jswent, please can you check that the name's I've added are how you would like to appear, and let me know if there are any other files that you've added.


josh-dev-things (approved):
Only change would be `Joshua James-Lee` instead of `Josh James-lee` if you could @Joseph-Edwards. Thank you!

PR #872, opened 2025-12-09
Minor updates to the doc-test pipeline

Joseph-Edwards:
This PR builds upon the changes introduced in #788. The main difference is in how we process the code blocks that should not be tested. Prior to this PR, these code blocks needed to include `\skip-test` to indicate that the test-generation script should skip them. Whilst this worked, that line also therefore appeared in the doc which isn't ideal.

This PR resolves this by introducing the aliases `\code_no_test` and `\end_code_no_test`. They should be used to enclose code blocks that should not be tested, but will be processed by doxygen in exactly the same way as `\code` and `\endcode` respectively.

## Before
image

## After
image

### DIFF
--- a/tests/test-docs-code-examples.cpp
+++ b/tests/test-docs-code-examples.cpp
@@ -429,8 +429,8 @@ namespace libsemigroups {
                           "./include/libsemigroups/ranges.hpp:87",
                           "[docs][quick]") {
     using rx::operator|;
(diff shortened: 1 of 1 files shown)


james-d-mitchell:
@Joseph-Edwards Formatting fails, but o/w happy to merge, will enable auto merge.

PR #886, opened 2026-01-17
Extract doc tests

james-d-mitchell:
Some updates related to extracting the doc tests so that it's easier to check that the file has been regenerated correctly (if necessary) on release.


james-d-mitchell:
@josh-dev-things if you can check you're okay with this too, I'll be most grateful

PR #889, opened 2026-02-02
sims: rm bad assert

james-d-mitchell:
This PR removes an invalid assertion from the hook function in `RepOrc`. @reiniscirpons could you possibly review?


reiniscirpons (approved):

james-d-mitchell:
I've updated the docs @reiniscirpons. Re: GAP we seem to always input an f.p. semigroup into `RepOrc` from GAP and properly handle any potential off by one issues, so I don't think there's any issue there.

PR #893, opened 2026-02-17
Use new clang-format action

Joseph-Edwards:
This PR switches from our usual clang-format action, to one that I have made. This is a temporary fix until https://github.com/jidicula/clang-format-action/issues/267 is resolved.


james-d-mitchell:
Nice thanks @Joseph-Edwards

PR #942, opened 2026-04-03
Remove redundant 'already finished' reporting

james-d-mitchell:
Drop the else branch in src/runner.cpp that reported "already finished, not running". This removes a noisy reporting path.

Resolves:

https://github.com/semigroups/Semigroups/issues/1169


james-d-mitchell:
The GAP semigroups package ci is failing because the most recent version of Orb has broken the semigroups package tests.

PR #945, opened 2026-04-04
cong: resolve issue #943

### LINKED ISSUE #943: Congruence runners added too early
The following:

~~~
    auto rg = ReportGuard(true);
    auto p  = presentation::examples::abacus_jones_monoid(6, 3);
    Presentation<word_type> f;
    f.alphabet(p.alphabet().size()).contains_empty_word(true);

    Congruence c(twosided, f);
    for (auto it = p.rules.begin(); it != p.rules.end(); it += 2) {
      congruence::add_generating_pair(c, *it, *(it + 1));
    }
    REQUIRE(c.number_of_classes() == POSITIVE_INFINITY);
~~~

is a `libsemigroups` version of the issue observed in Semigroups:

https://github.com/semigroups/Semigroups/issues/1167

The issue originates in the (it turns out, not correct) fix for 

https://github.com/libsemigroups/libsemigroups/issues/907

in 

https://github.com/libsemigroups/libsemigroups/pull/913

In this example in `Congruence::init` we correctly detect that `f` is infinite, then incorrectly deduce that the congruence must have infinitely many classes, and so no matter what generating pairs are added later no `ToddCoxeter` runners are added. This makes this example much slower than it is if `ToddCoxeter` would have been run. The fix will be to only add the runners just before `run_impl` is called for the first time. This way we have all the necessary information to know whether or not we should add `ToddCoxeter` runners.

james-d-mitchell:
This should resolve #943, initialising the runners just before the first run required a fair number of changes, which is why this is so large.

### DIFF
--- a/include/libsemigroups/cong-class.hpp
+++ b/include/libsemigroups/cong-class.hpp
@@ -34,6 +34,7 @@
 #include "presentation.hpp"        // for Presentation
 #include "todd-coxeter-class.hpp"  // for ToddCoxeter
 #include "types.hpp"               // for letter_type, wor...
+#include "word-graph-helpers.hpp"  // for complete_no_checks
 
 #include "detail/cong-common-class.hpp"  // for detail::CongruenceCommon
 #include "detail/race.hpp"               // for Race
@@ -159,9 +160,11 @@ namespace libsemigroups {
     // Congruence - data - private
     /////////////////////////////////////////////////////////////////////////
 
-    mutable detail::Race    _race;
-    mutable bool            _runners_initted;
-    std::vector<RunnerKind> _runner_kinds;
+    std::vector<Word>               _generating_pairs;
+    Presentation<Word>              _presentation;
+    mutable detail::Race            _race;
+    mutable bool                    _runners_initted;
+    mutable std::vector<RunnerKind> _runner_kinds;
 
     // Private init Congruence but not CongruenceCommon
     Congruence& init_no_base_classes();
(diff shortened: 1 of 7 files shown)


Joseph-Edwards:
The failing test in the Python bindings comes from:
```rst
.. doctest:: Python

    >>> from libsemigroups_pybind11 import (
    ...     congruence_kind,
    ...     Congruence,
    ...     WordGraph,
    ...     to,
    ... )

    >>> wg = WordGraph(7, [[1, 2], [1, 3], [4, 2], [5, 3], [4, 6], [5, 3], [4, 6]])
    >>> cong = to(congruence_kind.twosided, wg, rtype=(Congruence, list[int]))
    >>> cong.add_generating_pair([0], [1])
    <1-sided Congruence over <semigroup presentation with 2 letters, 0 rules, and length 0> with 1 gen. pair, 1 runners>    
    >>> cong.number_of_classes()
    1

    >>> cong = to(congruence_kind.twosided, wg, rtype=(Congruence, str))
    >>> cong.add_generating_pair("a", "b")
    <1-sided Congruence over <semigroup presentation with 2 letters, 0 rules, and length 0> with 1 gen. pair, 1 runners>    
    >>> cong.number_of_classes()
    1
```

We are trying to construct a two-sided congruence, and the returned object says it is a one-sided congruence. With the changes in this PR, this test returns a two-sided congruence which seems like the correct outcome to me. This might be indicative of a bug somewhere else.

Joseph-Edwards (approved):

james-d-mitchell:
Thanks @Joseph-Edwards, the "kind" of the congruence wasn't being set in the constructors previously, but now is:

https://github.com/libsemigroups/libsemigroups/pull/945/changes#diff-c22455c1e7354eca7850faad9b0647eb5a2f32ff4907389efb1e6d1271014e08R73

so I think this is resolved already, I also added tests for this to libsemigroups. I've also opened a PR to the python bindings with an update:

https://github.com/libsemigroups/libsemigroups_pybind11/pull/412

Issue #950, opened 2026-04-08
Matrix multiplication doesn't check the dimensions of the matrices


Joseph-Edwards:
Presently, we are able to multiply two matrices with dimensions that ought to be incompatible. For examples, we can multiply two matrices with dimension $2 \times 3$:

```
LIBSEMIGROUPS_TEST_CASE("Matrix", "028", "bad multiplication", "[quick]") {
  auto m = make<IntMat<2, 3>>({{1, 2, 3}, {4, 5, 6}});
  REQUIRE(m * m == make<IntMat<2, 3>>({{9, 12, 3}, {19, 26, 6}}));
}
```

We do have a warning in the doc that mentions something about this:
```
//! \warning
//! The matrices must be of the same dimensions, although this is not
//! verified.
```
but this is suspicious to me. Do we *really* want the matrices to have the same dimensions, or do we want them to have compatible dimensions, or do we want them to be the same dimension and square?

james-d-mitchell:
Looking at the code, it only claims to work for square matrices (this is asserted at the start), so I'll do at least two things to resolve this: 

1. write a `product_inplace` function that checks its args
2. update the documentation to indicate what only square matrices of equal dimensions are allowed.

How does that sound @Joseph-Edwards ?

Joseph-Edwards:
These suggestions sound reasonable to me. Will `operator*` use the `no_checks` or the `checks` version?

james-d-mitchell:
The `operator*` will use the checks version, and I've implemented things like `product_no_checks` etc, this seems consistent with what we're doing elsewhere, such as in `Meeter` and `Joiner` which have a call operator that does checks and a function `call_no_checks` which does not.

PR #954, opened 2026-04-09
matrix: impl `product_inplace` with checks

james-d-mitchell:
This is WIP, I need to add/update some of the doc I think still.

### DIFF
--- a/include/libsemigroups/matrix.hpp
+++ b/include/libsemigroups/matrix.hpp
@@ -6977,8 +6977,8 @@ namespace libsemigroups {
   //! compile-time, and those where it is not.  Both types can be accessed via
   //! the alias template \ref ProjMaxPlusMat : if \c R or \c C has value
   //! \c 0, then the dimensions can be set at run time, otherwise \c R and
-  //! \c C are the dimensions.  The default value of \c R is \c 0, \c C is \c R,
-  //! and `Scalar` is `int`.
+  //! \c C are the dimensions.  The default value of \c R is \c 0, \c C is
+  //! \c R, and `Scalar` is `int`.
   //!
   //! Matrices in both these classes are modified when constructed to be in a
   //! normal form which is obtained by subtracting the maximum finite entry in
--- a/include/libsemigroups/matrix.hpp
+++ b/include/libsemigroups/matrix.hpp
@@ -177,15 +177,21 @@ namespace libsemigroups {
     //! \tparam Mat the type of the argument, must satisfy \ref IsMatrix<Mat>.
     //!
     //! \param x the matrix to check.
+    //! \param arg_desc a string_view that describes the argument being checked
+    //! (defaults to `"the argument"`.
     //!
     //! \throws LibsemigroupsException if the number of rows in \p x does not
     //! equal the number of columns.
     template <typename Mat>
-    auto throw_if_not_square(Mat const& x) -> std::enable_if_t<IsMatrix<Mat>> {
+    auto throw_if_not_square(Mat const&       x,
+                             std::string_view arg_desc = "the argument")
+        -> std::enable_if_t<IsMatrix<Mat>> {
       if (x.number_of_rows() != x.number_of_cols()) {
-        LIBSEMIGROUPS_EXCEPTION("expected a square matrix, but found {}x{}",
-                                x.number_of_rows(),
-                                x.number_of_cols());
+        LIBSEMIGROUPS_EXCEPTION(
+            "expected {} to be a square matrix, but found a {}x{} matrix",
+            arg_desc,
+            x.number_of_rows(),
+            x.number_of_cols());
       }
     }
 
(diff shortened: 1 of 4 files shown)


james-d-mitchell:
Will resolve #950

james-d-mitchell:
I think this is good to go, @Joseph-Edwards any chance you could please review?

include/libsemigroups/matrix.hpp
```diff
@@ -177,15 +177,21 @@ namespace libsemigroups {
     //! \tparam Mat the type of the argument, must satisfy \ref IsMatrix<Mat>.
     //!
     //! \param x the matrix to check.
+    //! \param arg_desc a string_view that describes the argument being checked
+    //! (defaults to `\"the argument\"`.
```
Joseph-Edwards:
```suggestion
    //! (defaults to `"the argument"`.
```
We don't need to escape the quotes here:

Image

include/libsemigroups/matrix.hpp
```diff
@@ -198,20 +204,29 @@ namespace libsemigroups {
     //!
     //! \param x the first matrix to check.
     //! \param y the second matrix to check.
+    //! \param arg_desc_x a string_view that describes the argument \p x being
+    //!        checked (defaults to `"the 1st argument"` ).
```
Joseph-Edwards:
```suggestion
    //!        checked (defaults to `"the 1st argument"`).
```

include/libsemigroups/matrix.hpp
```diff
@@ -567,21 +625,36 @@ namespace libsemigroups {
...
       }
 
+      // not noexcept because product_inplace_no_checks isn't
+      Subclass operator*(Subclass const& y) const {
+        matrix::throw_if_bad_dim(static_cast<Subclass const&>(*this),
+                                 y,
+                                 "the 1st factor",
+                                 "the 2nd factor");
+        return product_no_checks(y);
+      }
```
Joseph-Edwards, line 657: Should this also call `throw_if_not_square` on `*this` and `y` as in `product_inplace`? As it stands, the following still passes:
```cpp
auto m = make<IntMat<2, 3>>({{1, 2, 3}, {4, 5, 6}});
REQUIRE(m * m == make<IntMat<2, 3>>({{9, 12, 3}, {19, 26, 6}}));
```

include/libsemigroups/matrix.hpp
```diff
@@ -890,21 +972,40 @@ namespace libsemigroups {
...
+      // TODO add tests
+      void operator+=(RowViewCommon const& x) {
+        throw_if_bad_dim(*this, x, "the 1st summand", "the 2nd summand");
+        plus_inplace_no_checks(x);
+      }
+
+      // not noexcept because operator+= isn't
+      Row plus_no_checks(RowViewCommon const& that) const {
+        Row result(*static_cast<Subclass const*>(this));
+        result += static_cast<Subclass const&>(that);
```
Joseph-Edwards:
```suggestion
        result.plus_inplace_no_checks(static_cast<Subclass const&>(that));
```

include/libsemigroups/matrix.hpp
```diff
@@ -2430,6 +2621,26 @@ namespace libsemigroups {
...
+    //! This function returns the product of two matrices.
+    //!
+    //! \param that the matrix to multiply by `this`.
+    //!
+    //! \returns The product of the two matrices.
+    //!
+    //! \throws LibsemigroupsException if either of the following apply:
+    //! * `this` does not point at a square matrix;
+    //! * \p x or \p y do not have the same dimensions as the matrix pointed at
+    //!   by `this`.
```
Joseph-Edwards:
```suggestion
    //! * the matrix pointed at by \p that does not have the same dimensions as
    //!   the matrix pointed at by `this`.
```

Joseph-Edwards (changes requested):
Some small doc changes, and the inclusion of some more checks in `operator*`, but then I would be happy to merge. Thanks @james-d-mitchell!

include/libsemigroups/matrix.hpp
james-d-mitchell: Good point yikes can't believe I missed that!

include/libsemigroups/matrix.hpp
james-d-mitchell: Good catch, although, we say "the matrix pointed to" for `this` because it's a pointer, whereas `that` is not, so I've adjusted this a little.

james-d-mitchell:
> Some small doc changes, and the inclusion of some more checks in `operator*`, but then I would be happy to merge. Thanks @james-d-mitchell!

Thanks @Joseph-Edwards I've applied your suggestions, and add an extra check in `operator*` for the matrices to be square, and added the integer matrix test case that prompted this PR in the first place. O/w not further changes.

Joseph-Edwards (approved):

PR #955, opened 2026-04-13
Split `matrix.hpp` into multiple files, out-of-line functions, and otherwise update

james-d-mitchell:


Joseph-Edwards:
There are some warnings coming from the various matrix files when the doc is being built. A larger number of these are resolved when I try building the doc with Doxygen v1.16.1. We're probably due a doxygen update anyway, so I think I'll try and update it, iron out any kinks, and fix these doc issues later. Therefore, if the CI passes, I'm happy for this to be merged. Thanks @james-d-mitchell!

PR #962, opened 2026-04-23
bitset: add tpp file

james-d-mitchell:


james-d-mitchell:
No idea why valgrind is suddenly failing. I think I'll rebase this onto `main` since it appears that the error is coming from `matrix.hpp` which was broken up in `main`

PR #963, opened 2026-04-23
Constant: stricter SFINAE conditions for implict conversion

Joseph-Edwards:
This PR resolves an issue we were facing in the Python bindings whereby some constants were incorrectly being implicitly converted. Instead of static-asserting that some conditions on `T` hold, we move these into the template parameter list and take advantage of SFINAE.


james-d-mitchell (approved):
Nice thanks @Joseph-Edwards

PR #964, opened 2026-04-24
Add wreath standardization for word graphs

jswent:
Also just realized I had this on my local branch from last semester. Definitely needs a review and probably a bit of work, but perhaps someone else can take it over. 

Taken from Sims Chapter 3


james-d-mitchell:
Included in #1034

PR #969, opened 2026-05-04
Support an external copy of magic_enum

orlitzky:
You are more and more likely to find magic_enum installed on the system:

* https://repology.org/project/magic-enum/versions

We have it on Gentoo, and it's pretty easy in this case, so I made a nice option for it:

1. Add a new `AX_CHECK_MAGIC_ENUM()` macro for `--with-external-magic-enum={yes,no}`
2. Run that macro in `./configure`
3. If `$with_external_magic_enum = yes`, skip adding `third_party/...` to the include path, and do not try to install any files from the bundled copy of magic_enum.

That's pretty much it, since in this case the code already uses

```C
#include <magic_enum/magic_enum.hpp>
```

which is how you include the copy from `/usr/include` anyway.

### DIFF
--- a/.github/workflows/config-options.yml
+++ b/.github/workflows/config-options.yml
@@ -51,7 +51,7 @@ jobs:
 
   external-deps:
     timeout-minutes: 45
-    name: ${{ matrix.external-backward }} ${{ matrix.external-fmt }} ${{ matrix.external-eigen }}
+    name: ${{ matrix.external-backward }} ${{ matrix.external-fmt }} ${{ matrix.external-eigen }} ${{ matrix.external-magic-enum }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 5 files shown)


james-d-mitchell:
Super @orlitzky thank you so very much! This was on our list of things to do, but we didn't get round to it yet, so this is amazing.

james-d-mitchell:
@orlitzky I added a ci job to check that the added code works, and the jobs have failed. Perhaps I've made a mistake, could you please have a look and see if anything stands out? I tried this locally, and it seems that not using `pkg-config` in `m4/ax_check_magic_enum.m4` which fails to locate `magic_enum` which is installed at: `/Users/jdm3/.local/share/mamba/envs/libsemigroups-dev/include`. If I do:

~~~
./configure CXX="ccache g++" CXXFLAGS="-O2 -g -I/Users/jdm3/.local/share/mamba/envs/libsemigroups-dev/include"  --disable-backward --with-external-magic-enum
~~~

then it works as expected. I don't exactly follow the rationale for not using pkg-config, could you possibly elaborate?

orlitzky:
Just being overly clever, I can change the macro to use pkg-config, it won't really be any more complicated. My rationale was that, whatever `--prefix` you use to install everything, the pkg-config file is going to give you `-I` flags that make `magic_enum/magic_enum.hpp` the right thing to include. But if you use a nonstandard prefix, you usually override `PKG_CONFIG_PATH`, `LDFLAGS`, `LD_LIBRARY_PATH`, and `CPPFLAGS` anyway to include the prefix... so `magic_enum/magic_enum.hpp` is what you'd include.

In the CI job...

> 2026-05-05T09:25:39.1190595Z   CXX: ccache g++
> 2026-05-05T09:25:39.1190850Z   CXXFLAGS: -fdiagnostics-color
> 2026-05-05T09:25:39.1191244Z   CONFIGFLAGS: --with-external-backward   --with-external-magic-enum
> 2026-05-05T09:25:39.1192056Z   PKG_CONFIG_PATH: /home/runner/micromamba/envs/libsemigroups/lib/pkgconfig:/home/runner/micromamba/envs/libsemigroups/share/pkgconfig/
> 2026-05-05T09:25:39.1192798Z   LD_LIBRARY_PATH: /home/runner/micromamba/envs/libsemigroups/lib
> 2026-05-05T09:25:39.1193307Z   BACKWARD_CPPFLAGS: -I/home/runner/micromamba/envs/libsemigroups/include

Both `PKG_CONFIG_PATH` and `LD_LIBRARY_PATH` are overridden, but the include dir is set only for backward (via `BACKWARD_CPPFLAGS`) rather than for all packages via `CPPFLAGS`. I think if it was set in `CPPFLAGS` it would work.

I'll rewrite the macro though. In the future it may be nice to have a version constraint on the accepted magic_enum.

orlitzky:
What do you know, we already need a version bound :)

I tried with magic_enum-0.9.6 and the path to the headers has changed.

orlitzky:
Testing with magic_enum-0.9.8 (released two days ago) while I'm at it, seems OK.

james-d-mitchell:
Super thanks so very much @orlitzky, I'll merge this once the CI has passed!

orlitzky:
No problem, I'm planning something similar for catch2 but that one is going to be more annoying. Thanks!

PR #970, opened 2026-05-06
Support an external catch library

orlitzky:
Catch is available on most Linux/BSD systems (though fewer ship 3.x):

* https://repology.org/project/catch/versions

Libsemigroups has already taken "Step 0" to migrate to catch-3.x, by shipping the `catch_amalgamated.{c,h}.pp` files from the upstream releases. Upstream however recommends building catch as a library, and (eventually) moving to piecewise headers:

* https://github.com/catchorg/Catch2/blob/devel/docs/migrate-v2-to-v3.md

This PR nudges things in that direction by adding support for a (linked) external catch-3.x library using the system `catch2/catch_all.hpp` in place of the bundled headers and compiled-in cpp file.

1. A new `--with-external-catch={yes,no}` flag is added to `./configure`. We use pkg-config to test for the 3.x library and headers. Despite what the docs say, I needed **both** Catch2 and Catch2Main.
2. To handle the two different header locations, we define a new constant `LIBSEMIGROUPS_CATCH_ALL_HEADER`. This goes in `config.hpp` so that other headers can still refer to it after being installed.
3. Tests and benchmarks are updated to include `config.hpp` and `LIBSEMIGROUPS_CATCH_ALL_HEADER`. In the tests this was accomplished by moving those includes into `test-main.hpp`.
4. In `Makefile.am`, we decide whether to compile the bundled `catch_amalgamated_patch.cpp` or link to the external library based on the variable `USE_BUNDLED_CATCH`.

Having to go through `config.hpp` is a bit ugly, but it should make it possible to update the bundled copy of catch without having to re-learn how anything works. For example, one approach I rejected would be to rename the bundled files to match the external names. Then no variable would be needed, but you'd have to remember to do that every time the bundled catch is upgraded.


james-d-mitchell:
Thanks so very much @orlitzky much appreciated. I think this approach is good, and very much appreciate the "not having to remember to do a thing when updating the bundled catch", this'll make my life much easier. If the CI passes I'll merge this, thanks again!

orlitzky:
Probably needs catch2 in environment.yml

orlitzky:
Thanks again!

PR #973, opened 2026-05-07
Add functions for working with commutators

Joseph-Edwards:
This PR adds the functions `commutator`, `add_commutator` and some other related functions. Some of the documentation was written with the help of codex.

### DIFF
--- a/include/libsemigroups/presentation.hpp
+++ b/include/libsemigroups/presentation.hpp
@@ -786,71 +786,17 @@ namespace libsemigroups {
     }
 
     template <typename Word>
-    void throw_if_contains_duplicates(Word const&      word,
-                                      std::string_view where) {
-      std::unordered_set<typename Word::value_type> letter_set;
-      for (auto const& letter : word) {
-        if (!letter_set.insert(letter).second) {
-          LIBSEMIGROUPS_EXCEPTION("invalid {}, the letter {} is duplicated!",
-                                  where,
-                                  detail::to_printable(letter));
-        }
-      }
-    }
+    void throw_if_contains_duplicates(Word const& word, std::string_view where);
 
     // TODO(1): This is very similar to the checks done in
     // throw_if_letter_not_in_alphabet, except there is no alphabet_map here. It
     // would be good if there was not duplication
     template <typename Word>
     void throw_if_word_not_over_alphabet(Word const& alphabet,
-                                         Word const& word) {
-      for (auto const& letter : word) {
-        if (auto it
-            = std::find(std::cbegin(alphabet), std::end(alphabet), letter);
-            it == std::cend(alphabet)) {
-          auto msg = fmt::format("invalid letter {}, valid letters are {}",
-                                 detail::to_printable(letter),
-                                 detail::to_printable(alphabet));
-          if constexpr (std::is_same_v<typename Word::value_type, char>) {
-            if (!std::isprint(letter) && detail::isprint(alphabet)) {
-              msg += fmt::format(
-                  " == {}",
-                  std::vector<int>(std::cbegin(alphabet), std::cend(alphabet)));
-            }
-          }
-          LIBSEMIGROUPS_EXCEPTION(msg);
-        }
-      }
-    }
+                                         Word const& word);
 
     template <typename Word>
-    void throw_if_bad_inverses(Word const& alphabet, Word const& inverses) {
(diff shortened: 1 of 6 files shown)


tests/test-presentation.cpp
```diff
@@ -3499,4 +3499,361 @@ namespace libsemigroups {
...
+      p.alphabet({0, 1});
+      REQUIRE_EXCEPTION_MSG(
+          presentation::commutator(p, W{}, W{}, W{0, 0}),
+          "invalid inverses, the letter (char with value) 0 is duplicated!");
+      REQUIRE_EXCEPTION_MSG(presentation::commutator(p, W{}, W{}, W{0, 2}),
+                            "invalid letter (char with value) 2, valid letters "
+                            "are (char values) [0, 1]");
+
+      p.alphabet({0, 1, 2});
+      REQUIRE_EXCEPTION_MSG(presentation::commutator(p, W{}, W{}, W{1, 2, 0}),
```
james-d-mitchell, line 3619: I'm not proposing a change here, but the usage `commutator` is a bit cryptic, it might have been better in retrospect to implement a simple `Commutator` class so we could do things like `Commutator().alphabet("ab").inverses("cd").get("ab", "b")` or something. As I said I'm not proposing we do this, this comment would apply to lots of the other functions in `presentation` too, just thinking out loud.

tests/test-presentation.cpp
```diff
@@ -3499,4 +3499,361 @@ namespace libsemigroups {
...
+      // some reason, the string comparison fails. The output should be:
+      // "the rules (char values) [0, 1] = "" (rule 1) and (char values) [0, 0]
+      // = "" (rule 0) yield the conflicting values (char with value) 1 != (char
+      // with value) 0 for the inverse of (char with value) 0, please use the 2-
+      // or 3- argument version of this function to explicitly specify the
+      // inverses"
+      //
+      // TODO(1): Better exception message here; the "2- or 3- argument
+      // version of this function" refers to <try_detect_inverses>.
+      REQUIRE_THROWS(presentation::commutator(p, W{}, W{}));
```
james-d-mitchell: Throws `LibsemigroupsException`?

james-d-mitchell: PS: I also can't see why `REQUIRE_EXCEPTION_MSG` doesn't seem to work here, the strings catch complains about do compare as equal, so this is likely an issue with Catch2 (maybe strings containing `\"`) or how we're using it.

include/libsemigroups/presentation.hpp
```diff
@@ -2053,6 +2116,59 @@ namespace libsemigroups {
...
+    //! \param p the presentation.
+    //! \param letters the word to contain the letters with inverses.
+    //! \param inverses the word to contain the inverses found.
+    //!
+    //! \throws LibsemigroupsException if
+    //! \ref Presentation::throw_if_bad_alphabet_or_rules throws.
+    //! \throws LibsemigroupsException if conflicting inverses for any letter
+    //! are detected.
+    template <typename Word>
+    void try_detect_inverses(Presentation<Word> const& p,
```
james-d-mitchell: Rename to `try_detect_group_inverses` and deprecate please.

include/libsemigroups/presentation.hpp
```diff
@@ -2053,6 +2116,59 @@ namespace libsemigroups {
...
+    //!
+    //! \tparam Word the type of the words in the presentation.
+    //! \param p the presentation.
+    //!
+    //! \throws LibsemigroupsException if
+    //! \ref Presentation::throw_if_bad_alphabet_or_rules throws.
+    //! \throws LibsemigroupsException if conflicting inverses for any letter
+    //! are detected.
+    template <typename Word>
+    std::pair<Word, Word> try_detect_inverses(Presentation<Word> const& p);
```
james-d-mitchell: Same comment, rename + deprecate

james-d-mitchell (changes requested):
Couple of minor issues, then let's merge it. Thanks @Joseph-Edwards, amazing how something so simple in principle grows arms and legs :)

Joseph-Edwards:
@james-d-mitchell, in case you were curious, there needs to be a blank line between the co-author thing and the commit summary:
> If you're using a text editor on the command line to type your commit message, ensure there is a blank line (two consecutive newlines) between the end of your commit description and the `Co-authored-by:` commit trailer.

PR #976, opened 2026-05-08
Fix weirdness in dark-mode

Joseph-Edwards:
This PR fixes the strange behaviour we were seeing in dark mode.


Joseph-Edwards:
@james-d-mitchell, please could you check this has fixed the issue you were facing. It seems to work on my machine

james-d-mitchell:
It does fix it, super thanks @Joseph-Edwards

PR #977, opened 2026-05-12
Add parsing for commutators in `operator""_p`

Joseph-Edwards:
This PR adds functionality for parsing commutators when using the `operator""_p` function. This is achieved by considering the comma to be the commutator binary operator, and assuming that the inverse of lowercase letters are the corresponding uppercase letter, and vice versa. This assumption is made clear in the doc.

It is not possible to use square brackets as one might want when specifying a commutator, but the additional work to match two different types of brackets didn't seem justified. One possible fix would be to treat all brackets as equal. This has the advantage of allowing strings like `"[a,b]"` to be parsed, but the disadvantage of also allowing strings like `"[a,b)"` to be parsed. In either case, this could be deferred to a later date.

### DIFF
--- a/src/word-range.cpp
+++ b/src/word-range.cpp
@@ -567,6 +567,18 @@ namespace libsemigroups {
     }
 
     namespace {
+      bool is_op(const char val) {
+        return std::string_view{",*^"}.find(val) != std::string_view::npos;
+      }
+
+      // Check op1 < op2 with respect to the order , < * < ^
+      bool compare_ops(const char op1, const char op2) {
+        LIBSEMIGROUPS_ASSERT(is_op(op1));
+        LIBSEMIGROUPS_ASSERT(is_op(op2));
+        std::string_view ops{",*^"};
+        return ops.find(op1) < ops.find(op2);
+      }
+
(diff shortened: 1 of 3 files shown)


james-d-mitchell:
This looks pretty amazing, thanks @Joseph-Edwards 

Can you add some tests for what happens if you forget the brackets `"ab,dc^10"` for example, what does that do?

src/word-range.cpp
```diff
@@ -656,6 +669,27 @@ namespace libsemigroups {
         return true;
       }
 
+      std::string get_unique_letters(std::string const& x) {
+        std::unordered_set<char> seen;
```
james-d-mitchell: This whole function can be replaced with 

~~~
std::unordered_set seen(x.begin(), x.end()); 
return std::string(seen.begin(), seen.end());
~~~

james-d-mitchell:
> ab,dc^10

I just tried this and it works as expected. Please just update the "get_unique" function and then let's merge, awesome!

james-d-mitchell (changes requested):
One small change then merge.

PR #978, opened 2026-05-12
ToddCoxeter: throw if non-core strategy and run_for

Joseph-Edwards:
This PR adds checks that `run_for` is not called on a `ToddCoxeter` object `tc` unless the strategy of `tc` is one of:
* `felsch`;
* `hlt`;
* `lookahead`; or
* `lookbehind`.

### DIFF
--- a/src/detail/todd-coxeter-impl.cpp
+++ b/src/detail/todd-coxeter-impl.cpp
@@ -818,9 +818,10 @@ namespace libsemigroups::detail {
                && strategy() != options::strategy::hlt
                && strategy() != options::strategy::lookahead
                && strategy() != options::strategy::lookbehind
-               && running_until()) {
-      LIBSEMIGROUPS_EXCEPTION("the strategy {} cannot be used with run_until !",
-                              strategy());
+               && (running_until() || running_for())) {
+      LIBSEMIGROUPS_EXCEPTION("the strategy {} cannot be used with {} !",
+                              strategy(),
+                              (running_until() ? "run_until" : "run_for"));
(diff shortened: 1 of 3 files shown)


Joseph-Edwards:
This is causing some issues in the following test:
```cpp
  LIBSEMIGROUPS_TEST_CASE("ToddCoxeter",
                          "033",
                          "symmetric_group(7) Moore_a",
                          "[todd-coxeter][quick][no-valgrind]") {
    auto rg = ReportGuard(false);

    size_t n = 7;
    auto   p = presentation::examples::symmetric_group_Moo97_a(n);

    ToddCoxeter tc(twosided, p);

    section_hlt(tc);
    section_felsch(tc);
    section_CR_style(tc);
    section_Cr_style(tc);
    section_R_over_C_style(tc);
    section_Rc_style(tc);

    tc.run_for(std::chrono::microseconds(1));
    ...
  }
```

This isn't really the behaviour we are trying to prevent by introducing this exception; in this case, the running is indeed stopped after 1 microsecond. Issues arise when running for a bit longer with the non-standard strategies, because it appears they are quite hard to interrupt once they get going.

I've altered this test so the `run_for` stuff happens before setting the strategies, and everything passes again.

Joseph-Edwards:
I've realised that this change may also cause a problem for `is_non_trivial`, because that repeatedly calls `run_for`. One possibility is to set the strategy to `hlt` at the start of this function, and then restore it at the end. Another option is to scrap the changes in this PR completely, though I think the first option is better. @james-d-mitchell, do you have any thoughts on this?

Issue #82, opened 2018-05-01
Add C-bindings


james-d-mitchell:
It'd be great to have C-bindings for `libsemigroups`, this would make it more straightforward to interface with other languages (such as, for example, Clojure @dersu).

PR #981, opened 2026-05-13
Initial C bindings

jfjameslee:
Making a start on adding C Bindings for libsemigroups (#82).

I will attempt to continue adding to this over the summer, however any VIP student next semester with understanding of C would be able to resume adding bindings.

### DIFF
--- a/.github/workflows/clang-format-check.yml
+++ b/.github/workflows/clang-format-check.yml
@@ -16,12 +16,14 @@ jobs:
           - src
           - tests
           - benchmarks
+          - c-bindings/include
+          - c-bindings/src
     steps:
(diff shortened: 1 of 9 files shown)


jswent:
Hey Josh, this looks great! Just one note on directory placement: afaik the standard is to have C bindings in the same library, so you would have a separate `include/libsemigroups-c` for the API headers and then the implementation lives in `src` (or `src/c`). A good example of this is [LLVM](https://github.com/llvm/llvm-project/tree/main/clang/include). 

This allows the same build system to be reused, and essentially gives users the C API "for free" when they install libsemigroups. But @james-d-mitchell may have a preference on this

PR #983, opened 2026-05-14
Add features from stable-3.5

Joseph-Edwards:
This PR adds the commits in `stable-3.5` that were missing from `main`


Joseph-Edwards:
This PR isn't doing what I hoped it would. Will close and open a new one.

PR #985, opened 2026-05-14
Some more functionality for `dot` and `WordGraph`s

james-d-mitchell:
This PR adds better drawing of word graphs with node and edge labels, and also some helpers for `FroidurePin` that produce the node/edge labels automatically.

### DIFF
--- a/include/libsemigroups/dot.hpp
+++ b/include/libsemigroups/dot.hpp
@@ -568,9 +568,7 @@ namespace libsemigroups {
     //! \param name the name.
     //!
     //! \returns A \c bool.
-    [[nodiscard]] bool is_node(std::string const& name) const {
-      return _nodes.count(name);
-    }
+    [[nodiscard]] bool is_node(std::string const& name) const;
 
     //! \brief Check if there is a node with name obtained from an object.
     //!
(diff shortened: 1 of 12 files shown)


james-d-mitchell:
Sorry I need to rename some things before this is ready, jumped the gun

james-d-mitchell:
The python bindings will fail, I've added an update at:

https://github.com/libsemigroups/libsemigroups_pybind11/pull/434

include/libsemigroups/dot.hpp
```diff
@@ -115,6 +115,17 @@ namespace libsemigroups {
   //! [Graphviz]: https://www.graphviz.org
   class Dot {
    public:
+    //! Enum to indicate the kind of attribute.
+    enum class Attr : uint8_t {
+      //! Value indicating that the value of an attribute should be enclosed in
+      //! double quotes.
+      string,
+
+      //! Value indicating that the value of an attribute should be enclosed in
```
Joseph-Edwards:
```suggestion
      //! Value indicating that the value of an attribute should not be enclosed in
```

Joseph-Edwards (commented):
One small change (which I will make now), but then I'm happy to merge

PR #988, opened 2026-05-16
words: fix CTAD for older libc++

jswent:
This patch addresses the issue with older libc++ versions not supporting iterator-pair deduction guides for unordered_set, which fixes the downstream error in https://github.com/JuliaPackaging/Yggdrasil/pull/13743


james-d-mitchell:
Aha super thanks @jswent I noticed this too, but you beat me to it!

PR #990, opened 2026-05-17
Release candidate for v3.6.1

james-d-mitchell:
I'm about to disappear onto the train for a number of hours, and cannot produce the test timings (on battery with my laptop), but o/w if the CI passes this is good to go.


jswent:
an hour later and CI still not done 😂

PR #992, opened 2026-05-22
presentation: `to_gap_string` use the same gen. names

james-d-mitchell:
Updates `to_gap_string` to use the same generator names in GAP as in libsemigroups. Also add a default value for the variable name (this because when using this in python especially, I almost always forget the second argument, or don't care what value it has).


james-d-mitchell:
The feature to add to the python bindings in this case is the default second argument for `to_gap_string`.

PR #1007, opened 2026-06-12
Codex/add pre commit format hooks

james-d-mitchell:
Finally fed up with clang-format ci jobs always failing so taking some proactive steps to avoid that. This PR is generated by Codex, I've checked that this works as intended. It might be tempting to add further things to the pre-commit hooks, like linting. 

Linting with cpplint takes approx. 22s though, so I'm not sure this is a good idea.


Joseph-Edwards:
I think this is a very good idea. Perhaps this is a good point to update higher versions of clang-format too?

james-d-mitchell:
I think I'd rather do that separately, because it would need to be done in the python bindings and in the GAP Semigroups package too.

PR #1008, opened 2026-06-12
Knuth bendix search

james-d-mitchell:


tests/test-action.cpp
```diff
@@ -781,6 +781,7 @@ namespace libsemigroups {
                           "021",
                           "constructors",
                           "[quick][no-valgrind]") {
+    auto         rg = ReportGuard(false);
```
Joseph-Edwards, line 784: Is it necessary to set the report guard to false in a tests for objects that don't produce any reporting? Or are we going for the blanket policy that all quick tests should have `ReportGuard(false)`.

If it's the latter case, is there any way of moving this into the test macro so we don't have to write it every time?

Joseph-Edwards (commented):
All looks good to me. One small question about the report guard stuff, but once that is answered I am happy to merge.

tests/test-action.cpp
james-d-mitchell: Good question. Here's my take: it isn't necessary, but there are ~2.5k test cases of which approximately ~2.2k are quick tests, and it doesn't seem doable to figure out which might report something and which not (maybe this is easy, but I doubt it). We could add something into a new macro that set this always to `true`, but given that'd require at least as many changes as the current version, I'm not really in favour of that either. Another thing to keep in mind is that when debugging (or other) we might want to enable the reporting to see what's going on, and it's quite convenient if the line of code required is already there. So, all in all, I think doing it like this is the least worst way to go.  Might note that there's also the script which Codex wrote to check that this is enforced correctly, which we could add as a pre-commit/push hook too perhaps?

Joseph-Edwards: Yeah that makes sense. I was hoping that our test macros were implemented in such a way that we could include something like `if "quick" in tags then add ReportGuard(false)` but, after looking at the implementation, that doesn't sound good.

Adding a pre-push hook seems like a good idea. I'll open an issue to add that.

Joseph-Edwards:
@james-d-mitchell are you happy for me to rebase this branch, squash the commits and then merge?

james-d-mitchell:
Yup thanks @Joseph-Edwards

PR #1013, opened 2026-06-17
Alphabet

james-d-mitchell:
This PR adds a new class `Alphabet` to `libsemigroups`.


include/libsemigroups/alphabet-class.hpp
```diff
@@ -0,0 +1,694 @@
...
+    void try_set_letters(decltype(_letters_map)& letters_map,
+                         native_word_type&       old_letters);
+    void throw_if_duplicate_letters(decltype(_letters_map)& letters_map) const;
+  };  // class Alphabet
+
+  ////////////////////////////////////////////////////////////////////////
+  // Validation
+  ////////////////////////////////////////////////////////////////////////
+
+  //! \ingroup presentations_group
```
Joseph-Edwards: Would this function, and subsequent ones, be better documented with `Alphabet`, by using the `\relates Alphabet` special command?
(https://www.doxygen.nl/manual/commands.html#cmdrelates)

Joseph-Edwards (commented):
One small doc question, but otherwise I am happy for this to be merged. Thanks @james-d-mitchell

include/libsemigroups/alphabet-class.hpp
james-d-mitchell: Good point, yes they would I'll up date this first thing tomorrow thanks @Joseph-Edwards

PR #1017, opened 2026-06-17
Add `Presentation` orientation searcher

Joseph-Edwards:
This PR adds a function that permutes the orientation of the rules in a presentation, and then checks whether an alphabet exists where those rules are correctly oriented with respect to rpo, and then checks confluence. If this is succeeds, then the presentation has a confluent and terminating rewriting system.


Joseph-Edwards:
I think the changes in this PR are probably obsolete in light of the stuff were adding with the ranges, but I won't close it just yet.

PR #1019, opened 2026-06-18
Add pipeline infrastructure supporting Knuth-Bendix searches (including morpho-completion)

james-d-mitchell:
Fix somethings I missed in the original PR, namely `to_human_readable` and usable copy/move etc constructors.


james-d-mitchell:
I realised I did something not so good with `throw_if_todo_populated` that impacts the usability of TietzeExplorer interactively. I'd like to undo that before merging this.

PR #1021, opened 2026-06-18
Improvements to `is_terminating`

Joseph-Edwards:
This PR adds improvements to `is_terminating`, makes it a member function of `RewritingSystemTrie` and uses the function to try and prevent against infinite loops of rewriting. Specifically, if a the left-hand-side of a rule is contained within the right-hand-side of a rule, `is_terminating_no_reduce` will now return `tril::FALSE`.

### DIFF
--- a/include/libsemigroups/detail/rewriting-system.tpp
+++ b/include/libsemigroups/detail/rewriting-system.tpp
@@ -444,6 +444,74 @@ namespace libsemigroups::detail {
   }
 
   template <typename ReductionOrder>
+  void RewritingSystemTrie<ReductionOrder>::rewrite(native_word_type& v) {
+    reduce();
+    rewrite_no_reduce(v);
+  }
+
+  template <typename ReductionOrder>
+  tril RewritingSystemTrie<
+      ReductionOrder>::is_length_non_increasing_no_reduce() noexcept {
+    if constexpr (order::is_length_non_increasing_v<ReductionOrder>) {
+      return tril::TRUE;
+    }
+
+    if (is_reduced() != tril::TRUE) {
+      return tril::unknown;
+    }
+
+    for (auto const& rule : rules()) {
+      if (rule.first.size() < rule.second.size()) {
+        return tril::FALSE;
+      }
+    }
+
+    return tril::TRUE;
+  }
+
+  template <typename ReductionOrder>
+  bool
+  RewritingSystemTrie<ReductionOrder>::is_length_non_increasing() noexcept {
+    if constexpr (order::is_length_non_increasing_v<ReductionOrder>) {
+      return true;
+    }
+
+    reduce();
+    return is_length_non_increasing_no_reduce() == tril::TRUE;
+  }
+
+  template <typename ReductionOrder>
+  tril
+  RewritingSystemTrie<ReductionOrder>::is_terminating_no_reduce() noexcept {
+    if constexpr (order::is_well_founded_v<ReductionOrder>) {
+      return tril::TRUE;
+    }
+    if (is_length_non_increasing_no_reduce() == tril::TRUE) {
+      return tril::TRUE;
+    }
+    for (auto const& rule : rules()) {
+      if (std::search(rule.second.begin(),
+                      rule.second.end(),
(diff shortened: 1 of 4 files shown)


include/libsemigroups/detail/rewriting-system.tpp
```diff
@@ -373,11 +380,99 @@ namespace libsemigroups::detail {
...
+
+  template <typename ReductionOrder>
+  tril RewritingSystemTrie<ReductionOrder>::is_terminating() noexcept {
+    if constexpr (order::is_well_founded_v<ReductionOrder>) {
+      set_cached_terminating(tril::TRUE);
+      return tril::TRUE;
+    }
+    reduce();
+    return is_terminating_no_reduce();
+  }
```
Joseph-Edwards: This function might run forever in scenarios where `is_terminating_no_reduce` would terminate. Would it be better to do something like:
```cpp
if (is_terminating_no_reduce() == tril::unknown){
    reduce();
}
return is_terminating_no_reduce();
```

include/libsemigroups/detail/rewriting-system.tpp
```diff
@@ -515,6 +608,11 @@ namespace libsemigroups::detail {
       // fmt::print("{}\n", to_printable(v));
       return;
     }
+
+    // If the rewriting system is not terminating, then rewriting may last
+    // forever.
+    LIBSEMIGROUPS_ASSERT(is_terminating_no_reduce() != tril::FALSE);
```
Joseph-Edwards: This isn't much help if we aren't in debug mode. I originally intended to make a `no_checks` version of the function, but then I was struggling to see its purpose. We are checking terminating here to try and prevent `reduce()` from running forever, but we would also probably want to use the `no_checks` version in `reduce`, so this wouldn't fix anything.

Because the problem of termination is only relevant when the reduction `ReturnFalse` (a rarity), then this assertion felt like an okay compromise.

A more extreme take would be to only use these functions and maintain a cached terminating value when debug mode is enabled.

Joseph-Edwards:
Before this is merged, once we have settled on some of the implementational details, we should also add this to `RewritingSytemSet`, and move common things into the base class

include/libsemigroups/detail/rewriting-system.tpp
```diff
@@ -373,11 +380,99 @@ namespace libsemigroups::detail {
...
+    }
+
+    tril result = tril::unknown;
+
+    if constexpr (order::is_well_founded_v<ReductionOrder>) {
+      result = tril::TRUE;
+    } else if (is_length_non_increasing_no_reduce() == tril::TRUE) {
+      result = tril::TRUE;
+    } else {
+      for (auto const& rule : rules()) {
```
james-d-mitchell: Can you please double check that pending rules are oriented correctly at all times? If not, then this doesn't work.

include/libsemigroups/detail/rewriting-system.tpp
james-d-mitchell: Yes, I think this is a good idea to make the change you suggest.

include/libsemigroups/detail/rewriting-system.tpp
james-d-mitchell: I don't think we want this assertion, since you're really asserting whether or not rewriting `v` will terminate, not whether we can tell if all rewrites will terminate.

james-d-mitchell (changes requested):
Some comments, and changes, also needs to be rebased, and move the implementation to the base class so that it works for all rewriting systems.

Joseph-Edwards:
> Some comments, and changes, also needs to be rebased, and move the implementation to the base class so that it works for all rewriting systems.

There is no `reduce` function in the base class. To have these functions implemented for each rewriting system, we will either have to have duplicate implementations for each, or have a `reduce` function in the base class (either as a virtual function or a function that does nothing). What do you reckon @james-d-mitchell?

james-d-mitchell:
Because this overlaps (the lines changed at least) with the changes in #1026 and #1031, we will finish this after those are merged.

PR #1023, opened 2026-06-25
Weight finding

Joseph-Edwards:
This PR adds a function that takes a presentation and returns a vector of weights for the presentation's alphabet such that the rules of the presentation are oriented with respect to weighted-len-lex.


Joseph-Edwards:
Still need to iron out how this integrates with our build system, and make the range object.

Joseph-Edwards:
Distcheck failed, but only on the "post setup ccache" step; the actually distcheck part of the job has passed.

PR #1024, opened 2026-06-29
RewritingSystem: Add `confluence_percentage`

Joseph-Edwards:
This PR adds a function that computes the proportion of critical pairs do not break confluence. There is quite a bit in common with the existing `confluence` functions; I have attempted to reduce duplication where it seemed easy to do so, but it might be worthwhile spending a bit more time to de-duplicate further.

### DIFF
--- a/include/libsemigroups/detail/rewriting-system.hpp
+++ b/include/libsemigroups/detail/rewriting-system.hpp
@@ -292,9 +292,7 @@ namespace libsemigroups {
                                    Iterator first2,
                                    Iterator last2);
 
-      [[nodiscard]] std::pair<size_t, size_t> confluence_ratio() override {
-        return {0, 0};
-      }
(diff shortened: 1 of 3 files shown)


james-d-mitchell (approved):

include/libsemigroups/detail/rewriting-system.hpp
```diff
@@ -137,6 +137,8 @@ namespace libsemigroups {
         return _confluence_known;
       }
 
+      [[nodiscard]] virtual std::pair<size_t, size_t> confluence_ratio() = 0;
```
james-d-mitchell: I suggest not having this as a virtual function, and just implementing it twice in each of the derived classes.

james-d-mitchell (changes requested):
De-virtualize and I'll be happy to merge this!

PR #1026, opened 2026-07-03
Rewriting system stores an Order object

Joseph-Edwards:
This PR changes the rewriting systems so that they store a copy of a reduction order object. With this change, it is possible to use the rewriting systems with stateful orders such as weighted len-lex.


include/libsemigroups/detail/rules.hpp
```diff
@@ -89,12 +89,17 @@ namespace libsemigroups {
...
+      if (order(rule->lhs(), rule->rhs())) {
         std::swap(rule->lhs(), rule->rhs());
       }
     }
 
+    template <typename ReductionOrder>
+    void reorder(Rule* rule) {
+      reorder(rule, ReductionOrder{});
+    }
+
```
Joseph-Edwards: We probably only need the first of these `reorder` functions, but the one parameter version seems quite useful. Should we keep the one parameter one despite the fact we currently don't use it, or delete it?

include/libsemigroups/detail/rewriting-system.hpp
```diff
@@ -399,6 +409,14 @@ namespace libsemigroups {
 
       [[nodiscard]] std::pair<size_t, size_t> confluence_ratio();
 
+      [[nodiscard]] ReductionOrder const& order() const noexcept {
```
james-d-mitchell: I was originally going to comment that maybe this should go in the base class, since the code is duplicated in `Set` and `Trie`, but the base class doesn't have the reduction order so let's leave it as is. Could you maybe add a comment to this effect?

include/libsemigroups/detail/rules.hpp
james-d-mitchell: Wouldn't it make more sense to just move this function into `RewritingSystem`? I think I moved it here because I could at that point, but now it makes more sense to move it back.

james-d-mitchell:
@Joseph-Edwards this should be complete pending any issues that the CI throws up

PR #1029, opened 2026-07-11
Orders using alphabets

james-d-mitchell:
This PR does a fairly large rewrite of the reduction orders to make them uniform, and usable with an alphabet. It's not finished yet, I still need to make the changes to `RewritingSystem` and `KnuthBendix`.


james-d-mitchell:
If the CI passes I think this is good to go, it should be squashed when merged, I tried initially to have decent commit messages, but that fell apart towards the end.

include/libsemigroups/order.tpp
```diff
@@ -21,7 +21,80 @@
...
+
+    template <typename Iterator>
+    void throw_if_incompat_weights(std::vector<size_t> const& weights,
+                                   Iterator                   first,
+                                   Iterator                   last) {
+      auto const it = std::find_if(first, last, [&weights](auto letter) {
+        return static_cast<size_t>(letter) >= weights.size();
+      });
+      if (it != last) {
+        LIBSEMIGROUPS_EXCEPTION("letter value out of bounds, expected value in "
```
Joseph-Edwards: I think it would be better if this exception message mentioned weights. Something like "letter value not compatible with weights, expected value in range [0, {}), found {} in position {}"?

include/libsemigroups/order.tpp
```diff
@@ -21,7 +21,80 @@
...
+    void throw_if_incompat_weights(Alphabet<Word> const&      alphabet,
+                                   std::vector<size_t> const& weights,
+                                   Iterator                   first,
+                                   Iterator                   last) {
+      auto const it
+          = std::find_if(first, last, [&alphabet, &weights](auto letter) {
+              return alphabet.index_no_checks(letter) >= weights.size();
+            });
+      if (it != last) {
+        LIBSEMIGROUPS_EXCEPTION("letter index out of bounds, expected index in "
```
Joseph-Edwards: Similar comment as above

Joseph-Edwards (commented):
I think this looks very good! I did scrutinise all 5000 additions, but the actual implementation of the order stuff seems good. A small comment on the exceptions below, but otherwise I reckon this is good to go

PR #1035, opened 2026-07-22
Rev orders

james-d-mitchell:
This PR introduces reverse (right to left) versions of all reduction orders in libsemigroups.


james-d-mitchell:
I still need to:

- [x] add documentation that the prefix `rev_` or `Rev` indicates that the words are read from right to left, and the absence indicates they are read from left to right.
- [x] rename `Wreath -> Wr`
- [x] swap `RPO` <-> `RevRPO`

james-d-mitchell:
The libsemigroups_pybind11 tests will have to be updated because we swapped RPO <-> RevRPO, and now the output of a test is different.

Joseph-Edwards (approved):
I have added a missing `Cmp` struct, fixed some inconsistencies in the (Rev)RPO doc, uncommented the maf tests that needed `RevWrCmp`, added some missing doc and rebased. It's now good to go.

PR #1040, opened 2026-07-23
word-graph: add to/from_disparse6_string

james-d-mitchell:


james-d-mitchell:
This was almost completely written by Codex, I checked that the tests make sense, and that the output disparse6 strings create the same graphs as those in the tests in GAP.

james-d-mitchell:
After some thought, disparse6 format doesn't record edge labels, which is essential for `WordGraph` serialisation, and so this is not really suitable for libsemigroups.

PR #1042, opened 2026-08-05
Add `KnuthBendixBacktrack`

Joseph-Edwards:
This PR introduces the `KnuthBendixBacktrack` class: an iterator for stepping through confluent rewriting systems for presentations.


Joseph-Edwards:
Still to do:
- [ ] move implementation into cpp files
- [ ] add more tests
- [ ] add a stopping condition
- [ ] better integration with the rest of the library

PR #1044, opened 2026-08-05
Add maximum rewriting depth

Joseph-Edwards:
This PR adds a setting to the rewriting systems that can be used to limit how many individual rewriting steps are allowed to occur before an exception is thrown. The default value is `POSITIVE_INFINITY`, so this should have no impact on existing test cases.

### DIFF
--- a/include/libsemigroups/detail/rewriting-system.tpp
+++ b/include/libsemigroups/detail/rewriting-system.tpp
@@ -290,7 +290,8 @@ namespace libsemigroups::detail {
     }
 
     // position of the start of the unread suffix of the input word
-    size_t pos = n - 1;
+    size_t pos                = n - 1;
+    size_t number_of_rewrites = 0;
 
     RuleLookup lookup;
 
@@ -303,6 +304,13 @@ namespace libsemigroups::detail {
         // lookup in _set_rules doesn't necessarily mean that we have found a
         // rule whose lhs is contained in [v.begin(), v.begin() + pos), that's
         // why we do the second check in the if-condition above.
+        ++number_of_rewrites;
+        if (number_of_rewrites
+            > RewritingSystemBase::settings().max_rewriting_depth) {
+          LIBSEMIGROUPS_EXCEPTION(
+              "rewriting has exceeded the maximum rewrite depth, {}",
+              RewritingSystemBase::settings().max_rewriting_depth);
+        }
         Rule const* rule = (*it).rule();
         LIBSEMIGROUPS_ASSERT(is_suffix(v.begin(),
                                        v.begin() + pos,
@@ -804,6 +812,7 @@ namespace libsemigroups::detail {
     //    }
 
     // NEW VERSION
+    size_t number_of_rewrites = 0;
     _trie_nodes_visited_indices.clear();
     index_type current = _rule_trie.root;
     _trie_nodes_visited_indices.push_back(current);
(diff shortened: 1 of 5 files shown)


include/libsemigroups/detail/rewriting-system.tpp
```diff
@@ -691,6 +700,13 @@ namespace libsemigroups::detail {
         _trie_nodes_visited_indices.push_back(current);
         pos++;
       } else {
+        ++number_of_rewrites;
+        if (number_of_rewrites
+            > RewritingSystemBase::settings().max_rewriting_depth) {
+          LIBSEMIGROUPS_EXCEPTION(
+              "rewriting has exceeded the maximum rewrite depth, {}",
```
james-d-mitchell:
```suggestion
              "rewriting has exceeded the maximum rewrite depth, settings().max_rewriting_depth = {}",
```

include/libsemigroups/detail/rewriting-system.tpp
```diff
@@ -212,6 +213,13 @@ namespace libsemigroups::detail {
         // lookup in _set_rules doesn't necessarily mean that we have found a
         // rule whose lhs is contained in [v.begin(), v.begin() + pos), that's
         // why we do the second check in the if-condition above.
+        ++number_of_rewrites;
+        if (number_of_rewrites
+            > RewritingSystemBase::settings().max_rewriting_depth) {
```
james-d-mitchell: Maybe DRY this out also

james-d-mitchell (commented):
Can you please add an example where the exceptions introduced in this PR are triggered by RPO in KnuthBendix (as discussed just now IRL)?

james-d-mitchell (commented):
I'm happy with this subject to the comments I made earlier, in particular, please add/expose one of the RPO tests which will trigger the exceptions in this PR to ensure that it's really the rewriting depth that is the issue there.

james-d-mitchell:
I found three actionable issues while reviewing this change:

1. **Preserve the depth limit across copies.** `RewritingSystemBase`'s copy/move assignment operators do not copy `_settings`. A copied system therefore gets `POSITIVE_INFINITY` for `max_rewriting_depth`, while assignment retains the destination's previous limit. This can silently remove the protection against runaway rewriting. Please copy/move `_settings` and add regression coverage.

2. **Avoid orphaning a pending rule when the limit is exceeded.** The new checks are inside `rewrite_no_reduce`, which `reduce()` calls after removing a rule with `pop_pending_rule()`. If normalization of that rule exceeds the limit, the exception exits before the rule is restored or made active/inactive, leaving the rewriting system partially mutated and missing that rule. This applies to both the set and trie implementations. Please make reduction exception-safe or enforce the limit outside this mutation window.

3. **Use a unique test-case number.** `RewritingSystem 019` is already used by the preceding test. The test runner emits a duplicate-number warning, and `tests/AGENTS.md` requires uniqueness within each family. `etc/test-next-number.sh test_rewriting_system` reports `020` as available.

Focused verification on the local branch:
- `test_rewriting_system [quick]`: 27 cases and 107 assertions passed, with the duplicate-number warning.
- `git diff --check` passed.

AI disclosure: OpenAI Codex reviewed the diff and ran the focused checks; I reviewed and posted its findings.

james-d-mitchell:
Number 3 is probably an artifact of my rebasing

Joseph-Edwards:
I've addressed most of these concerns, and now have a few questions:
1. Should the `Settings` struct have an `init` member function that restores the member to their default value, or is it okay to hard-code the default values into `RewritingSystem::init`?
2. What is the best way avoiding pending rule orphanage (if this is indeed necessary at all)? My thoughts are to wrap the relevant calls to `rewrite_no_reduce` in `try` `catch` blocks, and do something with the popped rule in the case that rewriting throws. Is there a sleeker way of doing this?

Joseph-Edwards:
I've added the `Settings` init function and, in 408fd84b34e22aa22a60303b852a3b3cc2a882eb, added some try-catch blocks around the code that might throw that re-adds the rule pointers to the pending rules in the case of an error. This fixed the memory leaks I was seeing in `KnuthBendix` test 038, but I still need to see if the all the address sanitizer and valgrind tests pass in the CI. I'm not certain this is the best way of solving the issue over something that uses smart pointers (or another solution entirely)

james-d-mitchell:
pybind11 expected to fail until we get update it for #1056

Joseph-Edwards:
I ran the quick and standard `KnuthBendix` tests and didn't see a change in run-time as a result of the added-try-catch code

james-d-mitchell (approved):
Looks good, and let's merge if the CI passes (except pybind11).

PR #1045, opened 2026-08-07
Add colours for ToddCoxeter phases in the output graph

Joseph-Edwards:
This PR adds to the `ToddCoxeter` output graph script by introducing coloured regions that represent the phases of `ToddCoxeter`:
image


james-d-mitchell (approved):
Beautiful

PR #1046, opened 2026-08-11
Create the Gilman graph from the trie

### LINKED ISSUE #1020: Refactor the `gilman_graph` code to use the trie
The Gilman-graph of a `KnuthBendix` object can be constructed using the rule trie. We should refactor the current implementation to use the rule trie.

Joseph-Edwards:
Closes #1020.

This PR adds code to construct the Gilman graph from a trie.

### DIFF
--- a/include/libsemigroups/detail/rewriting-system.hpp
+++ b/include/libsemigroups/detail/rewriting-system.hpp
@@ -600,6 +600,16 @@ namespace libsemigroups {
     using RewriteTrie [[deprecated]]     = RewritingSystemTrie<LenLexCmp>;
     using RewriteFromLeft [[deprecated]] = RewritingSystemSet<LenLexCmp>;
 
+    // Variable template that checks if a type has a trie member functions
+    // TODO (0): Remove this when there is a better way to determine which type
+    // of rewriter we are using.
+    template <typename, typename = void>
+    constexpr bool has_trie = false;
+
+    template <typename T>
+    constexpr bool has_trie<T, std::void_t<decltype(std::declval<T>().trie())>>
+        = true;
+
   }  // namespace detail
 }  // namespace libsemigroups
 
--- a/tests/test-knuth-bendix-from-froidure-pin.cpp
+++ b/tests/test-knuth-bendix-from-froidure-pin.cpp
@@ -408,7 +408,7 @@ namespace libsemigroups {
     auto copy   = kb.gilman_graph();
     auto source = copy.target(0, 2);
     copy.remove_label_no_checks(2);
-    REQUIRE(source == 34);
+
     REQUIRE(copy.out_degree() == 2);
     REQUIRE(copy.number_of_nodes() == 51);
     REQUIRE(word_graph::is_acyclic(copy, source));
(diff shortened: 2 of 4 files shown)


include/libsemigroups/detail/rewriting-system.hpp
```diff
@@ -570,6 +570,16 @@ namespace libsemigroups {
     using RewriteTrie [[deprecated]]     = RewritingSystemTrie<LenLexCmp>;
     using RewriteFromLeft [[deprecated]] = RewritingSystemSet<LenLexCmp>;
 
+    // Variable template that checks if a type has a trie member functions
+    // TODO (0): Remove this when there is a better way to determine which type
+    // of rewriter we are using.
```
james-d-mitchell, line 605: I think this is okay for now, in C++20 we can do better, but that's for later!

tests/test-knuth-bendix-from-froidure-pin.cpp
```diff
@@ -403,7 +403,8 @@ namespace libsemigroups {
     auto copy   = kb.gilman_graph();
     auto source = copy.target(0, 2);
     copy.remove_label_no_checks(2);
-    REQUIRE(source == 34);
+    // REQUIRE(source == 34);
```
james-d-mitchell: why this change?

tests/test-knuth-bendix-from-froidure-pin.cpp
```diff
@@ -416,7 +417,7 @@ namespace libsemigroups {
     std::sort(nrvec.begin(), nrvec.end());
     source = std::distance(nrvec.begin(),
                            std::find(nrvec.begin(), nrvec.end(), source));
-    REQUIRE(source == 28);
+    // REQUIRE(source == 28);
```
james-d-mitchell: And this one?

james-d-mitchell (commented):
Review finding:

[P2] `include/libsemigroups/detail/knuth-bendix-impl.tpp:740` should use a wider loop counter. `KnuthBendix` explicitly supports 256-letter alphabets when `char` is unsigned. With `out_degree == 256`, the `unsigned char` counter wraps from 255 to 0, so `gilman_graph()` never terminates in release builds. Debug builds instead fail the preceding assertion, which also unnecessarily rejects the valid 255-letter case. Please iterate with a wider type and cast when passing the value as a letter.

Verification: rebuilt `test_knuth_bendix`; `[quick]` passed 3,430 assertions in 479 test cases, `[standard]` passed 69 assertions in 19 test cases, and `git diff --check` passed.

AI disclosure: This review was prepared with OpenAI Codex, which inspected the diff and ran the relevant build and tests.

tests/test-knuth-bendix-from-froidure-pin.cpp
Joseph-Edwards: The node order in the Gilman graph is implementation specific. In the trie-based implementaion, nodes are numbered in breadth-first order from the root, whereas in the old implementation the numbering relates to the order of the rules in the `KnuthBendix` object. In both cases, the node **labels** should be the same.

james-d-mitchell: Okay, maybe just completely remove the lines then, rather than commenting them out.

Joseph-Edwards:
@james-d-mitchell I think this is good to go now.

james-d-mitchell:
@Joseph-Edwards can you please rebase? I just tried but wasn't able to (at least not without more effort) because the rebase is non-trivial, and I'm not sure which is the newer code.

PR #1059, opened 2026-08-23
Refactor alphabet validation

james-d-mitchell:
Make Alphabet the canonical letter validator, separate Presentation's empty-word policy, and deprecate redundant compatibility validators while preserving Knuth-Bendix's extra-letter checks.

AI disclosure: Codex assisted with implementation, refactoring, documentation, and tests.

### DIFF
--- a/include/libsemigroups/alphabet-class.tpp
+++ b/include/libsemigroups/alphabet-class.tpp
@@ -129,7 +129,9 @@ namespace libsemigroups {
   void
   Alphabet<Word>::throw_if_letter_not_in_alphabet(native_letter_type c) const {
     if (empty()) {
-      LIBSEMIGROUPS_EXCEPTION("there are no letters in the alphabet");
+      LIBSEMIGROUPS_EXCEPTION(
+          "invalid letter {}, there are no letters in the alphabet",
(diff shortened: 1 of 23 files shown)


james-d-mitchell:
Some issues with this are:

0. ~~the failing CI~~
1. ~~I don't really understand why `internal_presentation` needs to be checked in the `CongruenceCommon` [include/libsemigroups/detail/cong-common-class.tpp‎](https://github.com/libsemigroups/libsemigroups/pull/1059#diff-73b91053715eb3bbd3f0b6c82ba6ee63d3cc237ab2c962764ca9d44db32e69c4). Surely given that `KnuthBendixImpl` etc are not user facing, we don't require checks versions of any functions there, and we shouldn't be throwing exceptions there either because they won't make any sense (i.e. if an exception would be thrown this is our fault as developers not the fault of the user).~~
2. ~~we should add a test for `ToddCoxeter::current_index_of` which is apparently not covered.~~

james-d-mitchell:
Python bindings expected to fail

james-d-mitchell:
@codex review

Joseph-Edwards (approved):

PR #1068, opened 2026-08-28
Split word graph hpp

james-d-mitchell:
This also permitted removing the complication of declaring word graph helper functions twice, once as deprecated in `word-graph.hpp` and again in `word-graph-helpers.hpp` in the v4 namespace.  This doesn't strictly speaking preserve backwards compatibility because (according to Codex, and I agree):

* Eight helper templates gained a defaulted max_depth parameter: the reachability and spanning-tree checked/unchecked overloads. Ordinary calls still compile, but code taking their addresses with the old function types will not. See [word-graph-helpers.hpp (line 1311)](https://github.com/libsemigroups/libsemigroups/pull/1068/changes#diff-821aa40af7c0b85477877927fddb4ffe0104d0a910106129513ff70bf01a46feR1309-R1313) and [word-graph-helpers.hpp (line 1515)](https://github.com/libsemigroups/libsemigroups/pull/1068/changes#diff-821aa40af7c0b85477877927fddb4ffe0104d0a910106129513ff70bf01a46feR1514-R1518).
* word-graph-view.hpp previously included the umbrella word-graph.hpp; it now includes only word-graph-class.hpp. Users relying on that transitive inclusion to access helpers may fail to compile. See [word-graph-view.hpp (line 36)](https://github.com/libsemigroups/libsemigroups/pull/1068/changes#diff-d7b31541f9da27b67882bfd945fcb1c9957cf8b5d814935a564faf6b9756dd6dR36).
* Checked standardize now rejects out-of-bounds targets, whereas the old root implementation documented and implemented ignoring them. See [word-graph-helpers.hpp (line 1642)](https://github.com/libsemigroups/libsemigroups/pull/1068/changes#diff-821aa40af7c0b85477877927fddb4ffe0104d0a910106129513ff70bf01a46feR1643).
* Directly including the retained but empty word-graph.tpp no longer supplies implementations. That usage is unconventional, but it is another strict compatibility difference.

The first item is real, but unlikely to be any kind of issue, the remaining three could be considered as "bugs" in any client code rather than our issue, so I think this a-okay.


include/libsemigroups/word-graph-helpers.hpp
```diff
@@ -65,3867 +66,2772 @@
...
+    //! this function constructs a word graph to contain the result, forwards
+    //! this and the other arguments to the overload of `call_no_checks` with
+    //! `n + 1` parameters, then returns the word graph containing the result.
+    //!
+    //! \tparam Node the type of the nodes in the word graphs which are
+    //! parameters to this function.
+    //!
+    //! \tparam Args parameter pack for the remaining arguments.
+    //!
+    //! \param x the first word graph to join/me
```
Joseph-Edwards: Please could you add a comment explaining why these aliases are included.

Joseph-Edwards (commented):
This is excellent! Especially the net loss of almost 4000 lines. I've only very briefly skimmed the files with big changes because I assume it is just moving things around, but from what I saw it seems good!

I have a small question about file names. I was expecting the the new names of the files to be
1. `word-graph`,
2. `word-graph-class`,
3. `word-graph-helpers`,
4. `word-graph-view`,
5. `word-graph-view-class`, and
6. `word-graph-view-helpers`

where `word-graph` included `word-graph-class` and `word-graph-helpers`, and `word-graph-view` included `word-graph-view-class` and `word-graph-view-helpers`. There would be no file that includes `word-graph` and `word-graph-view` things.

We don't currently have a `word-graph-view-class` file. Is there a reason for not separating the files out further like this?

include/libsemigroups/word-graph-helpers.hpp
Joseph-Edwards: For some reason the diff show above doesn't seem to align with the lines I highlighted (at least not in my browser). For the avoidance of doubt, I am referring to these aliases at the bottom of the `word-graph-helpers.hpp` file:

```cpp
  namespace v4 {
    using ::libsemigroups::Joiner;
    using ::libsemigroups::make;
    using ::libsemigroups::Meeter;
    using ::libsemigroups::operator<<;
    using ::libsemigroups::to_human_readable_repr;
    using ::libsemigroups::to_input_string;
  }  // namespace v4
```

include/libsemigroups/word-graph-helpers.hpp
```diff
@@ -65,3867 +66,2772 @@
...
+  // namespace
+  // \ingroup word_graph_group
+  //
+  // \brief Namespace containing helper functions for the \ref WordGraph
+  // class.
+  //
+  // Defined in `word-graph-helpers.hpp`.
+  //
+  // \brief This namespace contains helper functions for the \ref WordGraph
+  // class.
```
Joseph-Edwards:
```suggestion
  //! \ingroup word_graph_group
  //!
  //! \brief Namespace containing helper functions for the \ref WordGraph
  //! class.
  //!
  //! Defined in `word-graph-helpers.hpp`.
  //!
  //! \brief This namespace contains helper functions for the \ref WordGraph
  //! class.
```

james-d-mitchell:
> This is excellent! Especially the net loss of almost 4000 lines. I've only very briefly skimmed the files with big changes because I assume it is just moving things around, but from what I saw it seems good!
> 
> I have a small question about file names. I was expecting the the new names of the files to be
> 
> 1. `word-graph`,
> 2. `word-graph-class`,
> 3. `word-graph-helpers`,
> 4. `word-graph-view`,
> 5. `word-graph-view-class`, and
> 6. `word-graph-view-helpers`
> 
> where `word-graph` included `word-graph-class` and `word-graph-helpers`, and `word-graph-view` included `word-graph-view-class` and `word-graph-view-helpers`. There would be no file that includes `word-graph` and `word-graph-view` things.
> 
> We don't currently have a `word-graph-view-class` file. Is there a reason for not separating the files out further like this?

Probably laziness is the cause of this, I will update to include these files too.

include/libsemigroups/word-graph-helpers.hpp
james-d-mitchell: This is accidental, will remove!

james-d-mitchell:
@Joseph-Edwards thanks for the comments! I've updated the PR, are you happy to merge if the CI passes?

Joseph-Edwards:
Yep. Merge away!

PR #1073, opened 2026-09-01
dev: check ReportGuard only in changed quick tests

james-d-mitchell:
## Summary

- add a --changed mode that checks staged, unstaged, and untracked source files
- add a pre-push hook that runs the checker only on changed C++ files under tests/
- preserve the existing full-repository and explicit-path modes

## Validation

- pre-commit configuration validation passed
- the hook passed on a compliant test file
- the hook correctly rejected a noncompliant test file
- pre-push hooks passed while pushing this branch
- git diff --check passed

## AI disclosure

Codex implemented and tested the changes and prepared this pull request.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -40,3 +40,11 @@ repos:
     hooks:
       - id: codespell
         stages: ["pre-push"]
+  - repo: local
+    hooks:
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
After a discussion with @Joseph-Edwards we decided to go in another direction with this.

PR #1075, opened 2026-09-02
Split `word-range.*`

Joseph-Edwards:
This PR splits `word-range.*` into `word-range-class.*` and `word-range-helpers.*`. I made these changes when trying to add alphabets to the word iterators, and was having some trouble with circular includes. It made sense for this to be its own PR.

I haven't changed any functionality, just moved things around.


james-d-mitchell:
I approve in principle, but we should probably discuss some of the details tomorrow am?

james-d-mitchell:
I'm not sure if we don't need to retain `word-range.cpp` for backwards compatibility, even if it's empty, can you please double check what we do elsewhere?

Joseph-Edwards:
> I'm not sure if we don't need to retain `word-range.cpp` for backwards compatibility, even if it's empty, can you please double check what we do elsewhere?

We do retain the cpp files for backwards compatibility. I have updated this

Joseph-Edwards:
The Python bindings failing would suggest that I did not correctly maintain backwards compatibility. I will investigate.

Joseph-Edwards:
> The Python bindings failing would suggest that I did not correctly maintain backwards compatibility. I will investigate.

Upon investigation, it seems that `paths.cpp` in libsemigroups_pybind11 was relying on transitive includes to get `ToString`. This is resolved in https://github.com/libsemigroups/libsemigroups_pybind11/pull/487.

PR #1076, opened 2026-09-03
More word graph cleanups

james-d-mitchell:


include/libsemigroups/exception.hpp
```diff
@@ -100,15 +98,15 @@ namespace libsemigroups {
 
   namespace detail {
     template <typename Iterator, typename Map, typename Ignore = ReturnFalse>
-    [[nodiscard]] std::pair<Iterator, size_t> find_duplicates(Iterator first,
-                                                              Iterator last,
-                                                              Map&     seen,
-                                                              Ignore&& ignore
-                                                              = Ignore{});
+    [[nodiscard]] std::pair<Iterator, uint64_t> find_duplicates(Iterator first,
```
Joseph-Edwards: Is the idea here that `uint64_t` always represents the same type across different systems? If this is the case, why don't we do this everywhere we see size_t?

In the implementation, the second entry in the pair represents some index, which feels like an appropriate place to use `size_t`.

include/libsemigroups/word-graph-view-class.hpp
```diff
@@ -931,7 +931,10 @@ namespace libsemigroups {
     //!
     //! \returns True if this and that have the same number of nodes, out
     //! degree, and range over nodes with identical values and targets.
-    // TODO(v4) should be a helper
+    // TODO(v4) should be a helper, can't currently deprecate it because this
+    // file is included in (e.g.) sims.cpp, making the compiler give deprecation
+    // warnings which there's no way to get rid of legitimately (i.e. without
+    // pragma suppressing them).
```
Joseph-Edwards, line 937: Should we add a similar note to `FroidurePin::equal_to_no_checks`?

Joseph-Edwards: Also, I've just tried deprecating this and only get warnings coming from the implementation of `WordGraphView::operator==`. I replaced the call of `equal_to_no_checks` with its implementation, and then I didn't get any compiler warnings.

include/libsemigroups/word-graph-view-helpers.hpp
```diff
@@ -1344,12 +1465,133 @@ namespace libsemigroups {
...
+    //! subword graph of those nodes reachable from \p source and \f$n\f$ is
+    //! the number of edges.
+    template <typename Node>
+    [[nodiscard]] std::vector<Node>
+    topological_sort(WordGraphView<Node> const& wgv, Node source);
 
   }  // namespace word_graph
 
+  //! \ingroup word_graph_group
+  //! Output the edges of a wordGraphView to a stream.
```
Joseph-Edwards:
```suggestion
  //! \brief Output the edges of a wordGraphView to a stream.
```

include/libsemigroups/word-graph-view-helpers.hpp
```diff
@@ -1344,12 +1465,133 @@ namespace libsemigroups {
...
+  //! \param os the ostream.
+  //! \param wgv the word graph view.
+  //!
+  //! \returns
+  //! The first parameter \p os.
+  //!
+  //! \exceptions
+  //! \no_libsemigroups_except
+  template <typename Node>
+  std::ostream& operator<<(std::ostream& os, WordGraphView<Node> const&);
```
Joseph-Edwards:
```suggestion
  std::ostream& operator<<(std::ostream& os, WordGraphView<Node> const& wgv);
```

So that the doc doesn't complain about `wgv` not being a parameter.

include/libsemigroups/word-graph-view-helpers.hpp
```diff
@@ -1344,12 +1465,133 @@ namespace libsemigroups {
...
+  //! \tparam Node the type of the nodes in \p wgv.
+  //!
+  //! \param wgv the word graph view to validate.
+  //!
+  //! \throws LibsemigroupsException if any of the following hold:
+  //! * the underlying WordGraph object of \p wgv has not been set;
+  //! * the values \ref WordGraphView::start_node() is strictly greater than
+  //! \ref WordGraphView::end_node();
+  //! * any target in the portion of the underlying WordGraph represented by
+  //! \ref wgv is greater than or equal to \ref WordGraphView::number_of_nodes.
```
Joseph-Edwards:
```suggestion
  //! \p wgv is greater than or equal to \ref WordGraphView::number_of_nodes.
```

include/libsemigroups/word-graph-helpers-deprecated.hpp
```diff
@@ -0,0 +1,975 @@
...
+        "Use the last_node_on_path iterator overload "
+        "instead.")]] [[nodiscard]] std::pair<Node, word_type::const_iterator>
+    last_node_on_path(WordGraph<Node> const& wg,
+                      Node                   source,
+                      word_type const&       w) {
+      return last_node_on_path(WordGraphView(wg), source, w.cbegin(), w.cend());
+    }
+
+    //! \copydoc last_node_on_path(WordGraph<Node> const&, Node, word_type
+    //! const&)
```
Joseph-Edwards:
```suggestion
    //!  const&)
```
Without the added space, doxygen attempts to search for `last_node_on_path(WordGraph<Node> const&, Node, word_typeconst&)`, which it fails to find.

An alternative solution would be to add a `no-lint-next-line` thing and put it all on one line.

include/libsemigroups/word-graph-helpers.hpp
```diff
@@ -1000,11 +1100,76 @@ namespace libsemigroups {
...
+    //!
+    //! \returns Whether or not \p wg is standardized according to \p cmp.
+    //!
+    //! \throws LibsemigroupsException if any target in \p wg is out of
+    //! bounds.
+    //!
+    //! \note If \p cmp corresponds to one of the orders in \ref Order for which
+    //! there is a bespoke salgorithm for checking standardization, then
+    //! \ref is_standardized(WordGraph&, Order) should be used instead of this
+    //! function for improved performance.
```
Joseph-Edwards:
```suggestion
```
This seems to have slipped through the cracks during our rebase

Joseph-Edwards (changes requested):
Some small comments, but overall this looks great. Thanks @james-d-mitchell!

include/libsemigroups/word-graph-helpers.hpp
james-d-mitchell: "salgorithm"

include/libsemigroups/exception.hpp
james-d-mitchell: I think I probably changed from size_t to uint64_t when doing a search and replace, my bad, I'll revert that

Joseph-Edwards: Thanks. There were a few things like this in e7ecbae8700742d92bab64bf7eb0d4fba3343ad9 I think.

include/libsemigroups/word-graph-view-class.hpp
james-d-mitchell: > Should we add a similar note to `FroidurePin::equal_to_no_checks`?

I don't see how this is related exactly, they do different things, the `FroidurePin` version checks if two words equal the same element of the semigroup, this one checks if two `WordGraph`s are equal, although I now see that maybe the naming convention is not great

PR #1078, opened 2026-09-03
presentation: improve perf of `remove_trivial_rules`

james-d-mitchell:
The doc claimed `remove_trivial_rules` had linear complexity, but it was quadratic, this PR resolves this.


james-d-mitchell:
Some of the tests running in valgrind are running for 10s of minutes, I don't think that's caused by the changes in this pr but will update them here anyway

james-d-mitchell:
Seems like the alpine failure is possibly related to:

https://github.com/jirutka/setup-alpine/issues/32

PR #1086, opened 2026-09-09
Add docs query CLI for agents

jswent:


james-d-mitchell:
Looks great to me thanks @jswent

PR #1092, opened 2026-09-15
Add generic orders for WordRange

Joseph-Edwards:
This PR introduces the class `v4::WordRange`: a class that allows for iterating over words in an arbitrary order. I originally tried to change the existing `WordRange` in-place, but the differences in the API were too significant to reasonably navigate with deprecation warnings. Other things added in this PR:
* a `has_alphabet` variable;
* the `order::is_specialization_of` struct;
* the `const_wio_iterator` for iterating through words of an arbitrary order;
* the `const_wi(s)lo_iterator` facades that allow for arbitrary alphabet orders; and
* the replacement of `StringRange` with `WordRange<std::string>`.

Unfortunately, the unconditional alphabet wrapping of the iterators in `WordRange` means that we see a somewhat significant performance hit when using either `LenLexCmp` or `LexCmp` over a normalised alphabet:
```
benchmark name                       samples       iterations    est run time
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
new WordRange LenLexCmp for length                                             
0 to 10                                        100             1     5.48225 s 
                                        53.3845 ms    53.0003 ms    54.1895 ms 
                                        2.71767 ms    1.57383 ms    4.37751 ms 
                                                                               
old WordRange LenLexCmp for length                                             
0 to 10                                        100             1     3.03167 s 
                                        30.8215 ms    30.5903 ms    31.3845 ms 
                                        1.72035 ms    638.605 us    3.13712 ms 
```

There are several ways I think we could proceed here:
1. Do nothing. These classes are only really used as small helpers in the test files, so arguably their performance is not that important. In a case where performance is important, it is still possible to use `const_wi(s)lo_iterator_imp` directly.
2. Do some partial-template-specialization-flavoured jiggling around of `WordRange` that allows us to conditionally set the contents of the iterator variant based on the template parameter `Word`. In the case that the `Word` was `word_type`, we could have the `impl` versions of the iterators in the variant too. If we wanted to do this, we would also need an additional means of setting the size of the alphabet independently of the order. This might complicate things, but shouldn't be too hard.
3. Do something similar to option 2, but instead of specifying the default `LenLexCmp` and setting the alphabet size in two different steps, we could detect that the alphabet stored in the comparator is normalised. In such a case, the variant would be set to an `_impl` version. This would require fewer additions to the `WordRange` API, but would rely on run-time checks rather than compile-time checks.
4. Have separate `WordRange` classes for when the words should be wrapped or not. This would require the most work.

I think my preference is option 1 in the short term, and then maybe option 2 or option 3 in the future if we actually notice a performance issue. I'm happy to hear other ideas!


james-d-mitchell:
Super thanks @Joseph-Edwards, before deciding what to do, it might be a good idea to understand where the difference in performance is coming from, and maybe try to address that. Can we discuss this am?

include/libsemigroups/detail/word-iterators.hpp
```diff
@@ -21,53 +21,79 @@
...
-      word_type   _last;
-      size_type   _number_letters;
+      Word                                          _current;
+      size_type                                     _index;
+      size_type                                     _upper_bound;
+      Word                                          _first;
+      Word                                          _last;
+      std::vector<Word>                             _frontier;
+      Alphabet<Word>                                _alphabet;
+      std::function<bool(const Word&, const Word&)> _cmp;
```
james-d-mitchell:
```suggestion
      std::function<bool(Word const&, Word const&)> _cmp;
```

include/libsemigroups/detail/word-iterators.hpp
```diff
@@ -80,81 +106,189 @@ namespace libsemigroups {
...
+      using value_type      = Word;
 
      private:
-      word_type _current;
-      size_type _index;
-      word_type _last;
-      size_type _number_letters;
+      // _external_current and _external_current_set are mutable because we
+      // defined the "state" of the iterator to be determined by the state of
+      // the date defined in const_wilo_iterator_impl. The _external things are
```
james-d-mitchell:
```suggestion
      // the data defined in const_wilo_iterator_impl. The _external things are
```

include/libsemigroups/detail/word-iterators.hpp
```diff
@@ -21,53 +21,79 @@
...
   namespace detail {
-    class const_wilo_iterator {
+    //////////////////////////////////////////////////////////////////////
+    // wio = words in order
+    //////////////////////////////////////////////////////////////////////
+
+    // There is no const_wio_iterator_impl because the comparator, which
+    // contains an alphabet, is baked into the implementation of operator++.
+    // Therefore, there is no need for a facade and an impl.
+    template <typename Word = word_type>
```
james-d-mitchell:
```suggestion
    template <typename Word>
```

include/libsemigroups/detail/word-iterators.hpp
```diff
@@ -80,81 +106,189 @@ namespace libsemigroups {
...
+      ~const_wilo_iterator();
 
-      const_wislo_iterator(size_type n, word_type&& first, word_type&& last);
+      const_wilo_iterator(Alphabet<Word> const& alphabet,
+                          size_type             upper_bound,
+                          Word const&           first,
+                          Word const&           last);
 
-      ~const_wislo_iterator();
+      // This is called reset for consistency with wio.
```
james-d-mitchell, line 168: Can you please expand this comment a bit? I.e. what would the alternate name be `init`? If so, how is `reset` different than init, if it isn't different, given that we are introducing wio here, maybe we should just be calling them `init`?

include/libsemigroups/detail/word-iterators.hpp
```diff
@@ -21,53 +21,79 @@
...
       using iterator_category = std::forward_iterator_tag;
 
      private:
-      word_type   _current;
-      size_type   _index;
-      letter_type _letter;
-      size_type   _upper_bound;
-      word_type   _last;
-      size_type   _number_letters;
+      Word                                          _current;
```
james-d-mitchell, line 64: Be best if these were given in alphabetical order, this is the convention elsewhere, possibly grouped by mutable/non-mutable.

include/libsemigroups/detail/word-iterators.tpp
```diff
@@ -0,0 +1,323 @@
...
+    const_wio_iterator<Word>::~const_wio_iterator() = default;
+
+    template <typename Word>
+    template <typename Cmp, typename>
+    const_wio_iterator<Word>::const_wio_iterator(
+        typename const_wio_iterator<Word>::size_type upper_bound,
+        Word const&                                  first,
+        Word const&                                  last,
+        Cmp&&                                        cmp)
+        : _current(),
```
james-d-mitchell: Alphabetize!

include/libsemigroups/detail/word-iterators.tpp
```diff
@@ -0,0 +1,323 @@
...
+
+        // Get the next word
+        do {
+          if (_frontier.empty()) {
+            _index = UNDEFINED;
+            break;
+          }
+          LIBSEMIGROUPS_ASSERT(!_frontier.empty());
+          std::pop_heap(_frontier.begin(), _frontier.end(), flipped_cmp);
+          _current = _frontier.back();
```
james-d-mitchell:
```suggestion
          _current = std::move(_frontier.back());
```

include/libsemigroups/detail/word-iterators.tpp
```diff
@@ -0,0 +1,323 @@
...
+
+          if (_current.size() < _upper_bound) {
+            // Update the frontier with words of length at most _upper_bound
+            for (letter_type a : _alphabet.letters()) {
+              Word new_word = _current;
+              new_word.push_back(a);
+              if (!_cmp(_last, new_word)) {
+                _frontier.push_back(new_word);
+                std::push_heap(_frontier.begin(), _frontier.end(), flipped_cmp);
+              }
```
james-d-mitchell: Would it be possible to use a `MultiView` here to avoid the copy in line 115, and then `_frontier.emplace_back(_current); _frontier.back().push_back(a);` in line 118. Alternatively if that doesn't work you could also move the `Word new_word;` to the outermost scope, and do `new_word.assign(_current.begin(), _current.end()); new_word.push_back(a);`

include/libsemigroups/detail/word-iterators.tpp
```diff
@@ -0,0 +1,323 @@
...
+          _current = _frontier.back();
+          _frontier.pop_back();
+
+          if (_current.size() < _upper_bound) {
+            // Update the frontier with words of length at most _upper_bound
+            for (letter_type a : _alphabet.letters()) {
+              Word new_word = _current;
+              new_word.push_back(a);
+              if (!_cmp(_last, new_word)) {
+                _frontier.push_back(new_word);
```
james-d-mitchell:
```suggestion
                _frontier.push_back(std::move(new_word));
```

include/libsemigroups/detail/word-iterators.tpp
```diff
@@ -0,0 +1,323 @@
...
+        } while (_cmp(_current, _first));
+        LIBSEMIGROUPS_ASSERT(_current.size() <= _upper_bound);
+        LIBSEMIGROUPS_ASSERT(!_cmp(_last, _current));
+      }
+      return *this;
+    }
+
+    template <typename Word>
+    void const_wio_iterator<Word>::swap(const_wio_iterator& that) {
+      std::swap(_index, that._index);
```
james-d-mitchell: Alphabetize

include/libsemigroups/detail/word-iterators.tpp
```diff
@@ -0,0 +1,323 @@
...
+      std::swap(_upper_bound, that._upper_bound);
+      std::swap(_first, that._first);
+      std::swap(_last, that._last);
+      std::swap(_frontier, that._frontier);
+      std::swap(_alphabet, that._alphabet);
+      std::swap(_cmp, that._cmp);
+      _current.swap(that._current);
+    }
+
+    // Assert that the forward iterator requirements are met
```
james-d-mitchell: Maybe delete these lines

include/libsemigroups/detail/word-iterators.tpp
```diff
@@ -0,0 +1,323 @@
...
+    template <typename Word>
+    void const_wilo_iterator<Word>::reset(
+        typename const_wilo_iterator<Word>::size_type upper_bound,
+        Word const&                                   first,
+        Word const&                                   last) {
+      _alphabet.throw_if_letter_not_in_alphabet(first.cbegin(), first.cend());
+      _alphabet.throw_if_letter_not_in_alphabet(last.cbegin(), last.cend());
+      const_wilo_iterator_impl::init(_alphabet.size(),
+                                     upper_bound,
+                                     word_type(citow(_alphabet, first.cbegin()),
```
james-d-mitchell: Be nicer not to have the `word_type` here and to just pass iterators?

include/libsemigroups/detail/word-iterators.hpp
Joseph-Edwards: It's different from an `init` function because there is no equivalent constructor that takes `upper_bound`, `first` and `last`. In particular, the constructors need an alphabet and `upper_bound` is not used by wislo. These functions all need the same signature so that they can be used inside a call to `std::visit`.

If the lack of an analogous constructor doesn't matter, then I will rename these functions.

include/libsemigroups/detail/word-iterators.tpp
Joseph-Edwards: I tried using `MultiView`, but I was having trouble constructing one over a `word_type`. I opted to add a data member to store this temporary `Word` in rather than to move it to the outerscope because `operator++` might be called hundreds of thousands of times. This is possibly bad for other reasons, so perhaps we should talk about this more?

include/libsemigroups/detail/word-iterators.tpp
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    template <typename Word>
+    const_wio_iterator<Word> const& const_wio_iterator<Word>::operator++() {
+      if (_index != UNDEFINED) {
+        ++_index;
+
+        // We need flipped_cmp so that we can maintain a min-heap, rather than a
+        // max-heap
+        auto const& flipped_cmp
```
james-d-mitchell:
```suggestion
        auto flipped_cmp
```

include/libsemigroups/detail/word-iterators.tpp
```diff
@@ -0,0 +1,295 @@
...
+            = [&cmp = this->_cmp](Word const& lhs, Word const& rhs) {
```
james-d-mitchell:
```suggestion
            = [this](Word const& lhs, Word const& rhs) {
```

include/libsemigroups/detail/word-iterators.tpp
```diff
@@ -0,0 +1,295 @@
...
+                return cmp(rhs, lhs);
```
james-d-mitchell:
```suggestion
                return _cmp(rhs, lhs);
```

This is equivalent to what you had before, and a little shorter.

include/libsemigroups/order.hpp
```diff
@@ -5868,6 +5867,63 @@ namespace libsemigroups {
   //! reduction order comparison types.
   // TODO these should not only be specialised for Default
   namespace order {
+    //! \brief Struct for checking if a template parameter is a specialization
```
james-d-mitchell: I think we discussed this before, but I'd prefer to not have these changes in here, i.e. the `order::is_specialization_of` stuff.

include/libsemigroups/word-range-class.hpp
```diff
@@ -107,18 +230,38 @@ namespace libsemigroups {
...
 
   //! \ingroup words_group
   //! \brief Returns a forward iterator pointing to one after the end of the
   //! range from \p first to \p last.
   //! \copydoc cend_wilo(size_t, size_t, word_type&&, word_type&&)
-  [[nodiscard]] detail::const_wilo_iterator cend_wilo(size_t n,
-                                                      size_t upper_bound,
-                                                      word_type const& first,
-                                                      word_type const& last);
+  [[deprecated]] [[nodiscard]] detail::const_wilo_iterator<word_type>
```
james-d-mitchell, line 245: Can't we just delete these? They are covered by the template below aren't they?

include/libsemigroups/word-range-class.tpp
```diff
@@ -0,0 +1,221 @@
...
+        return (*this | rx::count());
+      }
+    }
+
+    template <typename Word>
+    WordRange<Word>& WordRange<Word>::init() {
+      _current_valid = false;
+      _first         = {};
+      _last          = {};
+      _upper_bound   = 0;  // does nothing if the comparison order is lenlex
```
james-d-mitchell, line 144:
```suggestion
      _upper_bound   = 0;  // does nothing if the comparison order is not lex
```
Not sure about this.

include/libsemigroups/word-range-class.tpp
```diff
@@ -0,0 +1,221 @@
...
+      _visited       = 0;
+      order(LenLexCmp((Alphabet(Word{}))));
```
james-d-mitchell:
```suggestion
      order(LenLexCmp((Alphabet())));
```

include/libsemigroups/word-range-class.tpp
```diff
@@ -0,0 +1,221 @@
...
+    template <typename Word>
+    WordRange<Word>& WordRange<Word>::operator=(WordRange&&) = default;
+
+    template <typename Word>
+    WordRange<Word>::~WordRange() = default;
+
+    template <typename Word>
+    template <typename Cmp, typename>
+    WordRange<Word>& WordRange<Word>::order(Cmp&& cmp) {
+      if constexpr (order::is_specialization_of_v<Cmp, LenLexCmp>) {
```
james-d-mitchell:
```suggestion
      if constexpr (std::is_same_v<Cmp, LenLexCmp<>> || std::is_same_v<Cmp, LenLexCmp<Word>>) {
```

and same for the other one?

src/detail/word-iterators.cpp
```diff
@@ -1,6 +1,6 @@
 //
 // libsemigroups - C++ library for semigroups and monoids
-// Copyright (C) 2024-2025 Joseph Edwards
+// Copyright (C) 2025-2026 James D. Mitchell
 //
```
james-d-mitchell, line 4: Hmmm, not sure why this is being changed, was it wrong before?

tests/test-kambites.cpp
```diff
@@ -486,8 +486,10 @@ namespace libsemigroups {
     REQUIRE(k.number_of_classes() == POSITIVE_INFINITY);
     REQUIRE(number_of_words(3, 4, 16) == 21'523'320);
 
-    StringRange s;
-    s.alphabet("cab").first("aabc").last("aaabc");
+    v4::WordRange<std::string> s;
+    s.order(LenLexCmp(Alphabet<std::string>("cab")))
```
james-d-mitchell:
```suggestion
    s.order(LenLexCmp(Alphabet("cab"s)))
```

tests/test-kambites.cpp
```diff
@@ -661,9 +663,11 @@ namespace libsemigroups {
 
     REQUIRE(k.number_of_classes() == POSITIVE_INFINITY);
 
-    StringRange lhs;
-    lhs.alphabet("abcdefghijkl").first("a").last("bgdk");
-    StringRange rhs = lhs;
+    v4::WordRange<std::string> lhs;
+    lhs.order(LenLexCmp(Alphabet<std::string>("abcdefghijkl")))
```
james-d-mitchell:
```suggestion
    lhs.order(LenLexCmp(Alphabet("abcdefghijkl"s)))
```

tests/test-kambites.cpp
```diff
@@ -782,8 +786,8 @@ namespace libsemigroups {
 
     Kambites<TestType> k(twosided, p);
 
-    StringRange s;
-    s.alphabet("abcd").first("a").last("aaaa");
+    v4::WordRange<std::string> s;
+    s.order(LenLexCmp(Alphabet<std::string>("abcd"))).first("a").last("aaaa");
```
james-d-mitchell:
```suggestion
    s.order(LenLexCmp(Alphabet("abcd"s))).first("a").last("aaaa");
```

james-d-mitchell: Similar comments throughout this file by the looks of things

james-d-mitchell (commented):
A few tweaks and this'll be good to go

include/libsemigroups/word-range-class.hpp
Joseph-Edwards: Not quite. This overload takes a `size_t` as a parameter for the alphabet size, whereas the template below takes an `Alphabet<Word> const&`.

include/libsemigroups/word-range-class.tpp
Joseph-Edwards: I think it's correct as is. The `_upper_bound` data member gets used by both `wilo` and `wio`, but not by `wislo`.

src/detail/word-iterators.cpp
Joseph-Edwards: I think I just copy-pasted from one of the other empty files that we need to delete. I will fix this

include/libsemigroups/order.hpp
Joseph-Edwards: Right you are. I knew we had discussed this, but I couldn't remember the conclusion. I will remove.

include/libsemigroups/detail/word-iterators.tpp
```diff
@@ -94,10 +94,9 @@ namespace libsemigroups {
 
         // We need flipped_cmp so that we can maintain a min-heap, rather than a
         // max-heap
-        auto const& flipped_cmp
-            = [&cmp = this->_cmp](Word const& lhs, Word const& rhs) {
-                return cmp(rhs, lhs);
-              };
+        auto flipped_cmp = [this](Word const& lhs, Word const& rhs) {
+          return this->_cmp(rhs, lhs);
```
james-d-mitchell: Pretty sure the `this->` is not necessary

Issue #1094, opened 2026-09-16
`COUNT_TRUES_BLOCK` sometimes returns the wrong result


Joseph-Edwards:
The function template `COUNT_TRUES_BLOCK` in bitset.hpp is implemented as:
```cpp
template <typename T>
static inline size_t COUNT_TRUES_BLOCK(T block) {
  return __builtin_popcountl(block);
}
```
 and it implicitly converts `block` into an `unsigned long`. At the call site of this function, `block` has a type equal to one of the following:
* `uint_fast8_t`;
* `uint_fast16_t`
* `uint_fast32_t` or
* `uint64_t`.

On most systems, this is fine; `unsigned long` is usually 64-bits, and the above types are at most 64-bit. However, on Windows, `unsigned long` is 32-bits, so `COUNT_TRUES_BLOCK(uint64_t)` will incorrectly truncate `block` to 32-bits, and hence the wrong value is returned.

The simplest way to fix this would be to replace `__builtin_popcountl` with `__builtin_popcountll` (note the extra 'l') which works with `unsigned long long`s. In this case, no truncation would take place, but this is perhaps less efficient when the actual data type uses fewer bits. An alternative solution would be to have a `constexpr if` that dispatches to the correct `__builtin_popcount` based on `sizeof(T)`. This would presumably be more efficient, but some care would need to be taken to ensure all of the necessary built-in functions were available.

I've tested both of these approaches on windows and both seem to work, but it's not clear to me which solution is better (specifically because of the awkwardness of checking whether all three built-ins are available in the second solution) so I'll hold off on making a PR until we've discussed this further.

This issue is the only thing that is currently preventing all of the quick tests from passing on Windows!

PR #1097, opened 2026-09-21
Pop count: always use ll

james-d-mitchell:
This PR should fix issue #1094. According to codex, there's no need for more code paths related to the size of various integers etc on the host machine because the same code is generated using the `ll` variant of the built in pop count. 

Assisted-by: Codex OpenAI <codex@openai.com>


james-d-mitchell:
@Joseph-Edwards if the CI passes would you be willing to have look?

Joseph-Edwards (approved):
I've checked these changes on on Windows and everything seems to pass now. Thanks @james-d-mitchell!

